PR #12, opened 2021-05-24
Process/scott pd

ayyubibrahimi:
Please review. 

Disregard the changes to baton_rouge_so_cprr.py as I made them before I saw the issue that you created. 

Thanks!

### DIFF
--- a/clean/scott_pd_cprr.py
+++ b/clean/scott_pd_cprr.py
@@ -76,3 +76,4 @@ if __name__ == "__main__":
     df.to_csv(
         data_file_path("clean/cprr_scott_pd_2020.csv"),
         index=False)
+
--- a/clean/scott_pd_cprr.py
+++ b/clean/scott_pd_cprr.py
@@ -2,7 +2,7 @@ from lib.columns import clean_column_names
 from lib.uid import gen_uid
 from lib.path import data_file_path, ensure_data_dir
 from lib.clean import (
-    clean_names, standardize_desc_cols, clean_dates, clean_sexes, clean_races, clean_datetimes
+    clean_names, standardize_desc_cols, clean_dates
 )
 import pandas as pd
 import sys
@@ -16,6 +16,21 @@ def split_name(df):
     df = df.drop(columns="full_name")
     return df
 
+
+def split_disposition_action(df):
+    outcomes = df["disposition/action"].str.split("/", expand=True)
+    df.loc[:, "disposition"] = outcomes.loc[:, 0].fillna("")
+    df.loc[:, "action"] = outcomes.loc[:, 1].fillna("")
+    df = df.drop(columns="disposition/action")
+    return df
+
+def clean_rank(df):
+    df.loc[:, "rank_desc"] = df.rank_desc.str.lower().str.strip() \
+        .str.replace("pofc", "officer", regex=False)\
+        .str.replace("lt.", "lieutenant ", regex=False)
+    return df
+
+
 def assign_agency(df):
     df.loc[:, "agency"] = "Scott PD"
     return df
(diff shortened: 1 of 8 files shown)


clean/scott_pd_cprr.py
```diff
@@ -0,0 +1,76 @@
...
+
+def clean_rank(df):
+    df.loc[:, "rank_desc"] = df.rank_desc.str.lower().str.strip() \
+        .str.replace("pofc", "officer", regex=False)\
+        .str.replace("lt.", "lieutenant ", regex=False)
+    return df
+
+def assign_agency(df):
+    df.loc[:, "agency"] = "Scott PD"
+    return df
```
pckhoi, line 34: checkout `columns.set_values`. You can do stuffs like this without defining an extra function. E.g.
```python
        ...
        .pipe(set_values, {
            'data_production_year': 2020,
            'agency': 'Brusly PD'
        })
```

clean/scott_pd_cprr.py
```diff
@@ -0,0 +1,76 @@
...
+        .str.replace("conduct unbecoming/rudeness", "conduct unbecoming (rudeness)", regex=False)
+    return df
+
+def clean():
+    df = pd.read_csv(data_file_path(
+        "scott_pd/scott_pd_cprr_2020.csv"))
+    df = clean_column_names(df) \
+        .rename(columns={
+        'rule_violation': 'charges'})
+    df.columns = ['notification_date', 'full_name', 'charges', 'appeal', 'disposition/action']
```
pckhoi, line 53: I guess you are trying to remove columns? In newer scripts I have stopped doing this as it obscure what columns are removed. We should try to use `drop(columns=[...])` for that purpose instead.

clean/scott_pd_cprr.py
```diff
@@ -0,0 +1,76 @@
...
+        .pipe(gen_uid, ["agency", "first_name", "last_name"])\
+        .pipe(gen_uid, ["agency", "uid", "notification_year", "notification_month", "notification_day"], "complaint_uid")
+    return df
+
+if __name__ == "__main__":
+    df = clean()
+    ensure_data_dir("clean")
+    df.to_csv(
+        data_file_path("clean/cprr_scott_pd_2020.csv"),
+        index=False)
```
pckhoi, line 76: All files should have one empty line at the end. Are you familiar with editorconfig? Perhaps we need to create that file to enforce this.

pckhoi:
I guess this isn't ready for review?

clean/scott_pd_cprr.py
ayyubibrahimi: Thanks.

clean/scott_pd_cprr.py
ayyubibrahimi: No. I will look into it now

ayyubibrahimi:
Fixed errors and created fuse script, edited Make file, and added to fuse/all.py and .wrglsync.yaml.

data/.wrglsync.yaml
```diff
@@ -72,6 +74,8 @@ repos:
     fileName: match/cprr_plaquemines_so_2019.csv
   match-lprr-louisiana-state-csc-1991-2020:
     fileName: match/lprr_louisiana_state_csc_1991_2020.csv
+  match-cprr-scott-pd-2020:
+    fileName: match/cprr_scott_pd_2020.csv
```
pckhoi, line 78: Going forward perhaps we don't need to keep clean and match script in WRGL. Doesn't look like anybody is interested in how those files change.

rajivsinclair: Let's discuss. Why do you think we should not? I think it's helpful to keep it all in one place.

pckhoi: Maybe I'm wrong but I just feel like we don't have any eyeballs on those files. But we can keep them if you're in doubt.

pckhoi:
@ayyubibrahimi in case you didn't know all PRs need to have one of these labels to be merged: "bug", "documentation", "enhancement".

pckhoi:
@ayyubibrahimi mark this as "Ready for review" when you think it's ready. Which I guess it is?

fuse/scott_pd.py
```diff
@@ -0,0 +1,61 @@
...
+        events.OFFICER_HIRE: {'prefix': 'hire', 'keep': ['uid', 'agency']},
+    }, ['uid'])
+    return builder.to_frame()
+
+
+if __name__ == '__main__':
+    cprr = pd.read_csv(
+        data_file_path("match/cprr_scott_pd_2020.csv"))
+    post = prepare_post()
+    per = rearrange_personnel_columns(post)
```
pckhoi, line 45: If CPRR contains officers not found in POST then you should include them. To fuse personnel data from 2 or more datasets look at `lib.personnel.fuse_personnel`

ayyubibrahimi: Fused.

pckhoi (approved):
LGTM

PR #14, opened 2021-05-25
Fix some errors for Baton Rouge SO and Scott PD

pckhoi:
Changes:
- Makefile: Correct file names, please make sure that `make` run without error.
- clean/baton_rouge_so_cprr.py:
    - Change column reference "full_name" to "name".
    - Make sure each top level function are 2-lines apart (Flake8)
    - Make sure no line is longer than 120 characters (Flake8)
- clean/scott_pd_cprr.py:
    - Remove `df.columns = ...` line. Please use `df.rename(columns={...` instead if you want to rename columns.
    - Drop empty rows. `complaint_uid` and `uid` were being generated for empty rows.
- match/scott_pd.py:
    - Replace `lib.match` with `datamatch`.
    - Correct `ThresholdMatcher` syntax to datamatch version 0.1.0
- scripts/download_input.py:
    - Replace stale link for Baton Rouge SO
    - Add `dl=1` to Scott PD link so that it could be downloaded.


ayyubibrahimi (approved):
Got it.

Issue #4, opened 2021-05-22
Document existing code in `lib` package


pckhoi:
Now that there are more developers working on this. This project should start having documentation at every level so that new people can be initiated into it quickly. One of the first area that we should tackle is to add docstring to all functions and classes found in `lib` package.

All docstrings should follow [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

PR #16, opened 2021-05-25
Add docstring to everything under lib

pckhoi:
Add docstring as per #4


pckhoi:
@ayyubibrahimi after this is merged I think it's probably a good idea for you to read up docstring for everything under `lib`. That way you will be familiar with all the existing tools (and avoid creating anything that already exist).

ayyubibrahimi (approved):
Got it. Thanks.

PR #23, opened 2021-06-09
Enhancement/new orleans da

ayyubibrahimi:
Hey Khoi,

There are a few issues that I can't figure out. Please see see below:

clean_new_orleans_da_cprr_brady_2021.py:
I can't replace/clean the text below because, seemingly, of the accents.

rule 2: moral conduct ; paragraph 0: adhÃ©rence to law
rule 2:moral conduct ; paragraph 01 adhÃ©rence to law

match/new_orleans_da.py
I'm finding no matches with per_new_orleans_so, but there are seemingly a few. 

fuse/new_orleans_da.py
1) should I not have created a fuse script? I only realized that you didn't create one for the baton_rouge_brady list after adding said files to the Make and wrgl.yaml file. 
2) I can't figure out how to make the event_uid unique. 

Thank you!


Makefile
```diff
@@ -11,6 +11,7 @@ all: data/fuse/per_port_allen_pd.csv data/fuse/event_port_allen_pd.csv data/fuse
 all: data/fuse/per_madisonville_pd.csv data/fuse/event_madisonville_pd.csv data/fuse/com_madisonville_pd.csv
 all: data/fuse/per_greenwood_pd.csv data/fuse/com_greenwood_pd.csv data/fuse/event_greenwood_pd.csv
 all: data/fuse/com_new_orleans_pd.csv data/fuse/uof_new_orleans_pd.csv data/fuse/per_new_orleans_pd.csv data/fuse/event_new_orleans_pd.csv
+all: data/match/per_new_orleans_da.csv data/fuse/event_new_orleans_da.csv data/fuse/com_new_orleans_da.csv
```
pckhoi, line 14: Doesn't look like there are any rule to make `data/match/per_new_orleans_da.csv`

clean/new_orleans_da_cprr_2021_brady.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def drop_rows_without_last_name(df):
+    df = df[df.last_name != 'test']
+    return df.dropna(subset=['last_name']).reset_index(drop=True)
+
+
+def clean():
+    df = pd.read_csv(data_file_path(
+        'new_orleans_da/new_orleans_da_cprr_2021.csv'))
```
pckhoi, line 311: The file newly added in download_input is `new_orleans_da/new_orleans_da_brady_list_2021.csv`. The name is wrong.

clean/new_orleans_da_cprr_2021_brady.py
```diff
@@ -0,0 +1,347 @@
...
+        .str.replace('rule 2: moral conduci;paragraph 01 adherence to law',
+                     'rule 2: moral conduct;paragraph 01 adherence to law', regex=False)\
+        .str.replace('rule 2 moral conduct ; paragraph 0 honesty and truthfulness',
+                     'rule 2: moral conduct;paragraph 03 honesty and truthfulness', regex=False)\
+        .str.replace('rule: 2: moral conduct;paragraph 01 adherence to law',
+                     'rule 2: moral conduct;paragraph 01 adherence to law', regex=False)\
+        .str.replace('rule 6: official info ; paragraph 02 false or inaccurate reports',
+                     'rule 6: official info;paragraph 02 false or inaccurate reports', regex=False)\
+        .str.replace(r'^rule 2: moral conduct ; paragraph 01 adhÃ©rence to law$',
+                     'rule 2: moral conduct ; paragraph 01 adherence to law')
```
pckhoi, line 143: These lines look like they could be condensed to just a few lines with proper use of regex.

clean/new_orleans_da_cprr_2021_brady.py
```diff
@@ -0,0 +1,347 @@
...
+        .str.replace('sustained dismissed under another investigation',
+                     'sustained | dismissed under another investigation', regex=False)\
+        .str.replace('sustained pending superintendent approval',
+                     'sustained | pending superintendent approval', regex=False)\
+        .str.replace('sustained pending suspension served',
+                     'sustained | pending suspension served', regex=False)\
+        .str.replace('sustained | dismissedl',
+                     'sustained | dismissed', regex=False)\
+        .str.replace('sustained | dismissedl (1 po dismissed)',
+                     'sustained | dismissed (1 po dismissed)', regex=False)\
```
pckhoi, line 177: Same here, please use regex to eliminate duplications.

match/new_orleans_da.py
```diff
@@ -0,0 +1,129 @@
...
+        'last_name': JaroWinklerSimilarity(),
+        'first_name': JaroWinklerSimilarity(),
+    }, dfa, dfb)
+    decision = 1
+    matcher.save_pairs_to_excel(data_file_path(
+        'match/new_orleans_da_cprr_2021_v_post_noso_pprr_2020_11_06.xlsx'), decision)
+    matches = matcher.get_index_pairs_within_thresholds(lower_bound=decision)
+    match_dict = dict(matches)
+
+    cprr.loc[:, 'uid'] = cprr.uid.map(lambda x: match_dict.get(x, x))
```
pckhoi, line 38: Why do you assign uid from post to cprr? What is the purpose?

match/new_orleans_da.py
```diff
@@ -0,0 +1,129 @@
...
+    df = float_to_int_str(
+        df, ['first_name', 'last_name'])
+    df = match_against_new_orleans_pd_personnel(df, pprr)
+    df = match_against_new_orleans_so_personnel(df, so_per)
+    post_noso = prepare_new_orleans_so_post_data()
+    post_nopd = prepare_new_orleans_pd_post_data()
+    post_event = pd.concat([
+        match_cprr_and_noso_post(df, post_noso),
+        match_cprr_and_nopd_post(df, post_nopd)
+    ]).drop_duplicates(ignore_index=True)
```
pckhoi, line 126: I don't think you understand what post_event is. It is events data extracted from POST. Please see this PR: https://github.com/ppact/processing/pull/22

pckhoi (commented):
What is the purpose of fusing a brady list with something else? You need to ask that question before writing a fuse script. I didn't write a fuse script for brady list because it is just a list of officers to be cooked down to just a list of uid. If you can match the uid then it's already enough.

Please take it slow Ayyub, we can do one on one video call on this if you want.

match/new_orleans_da.py
ayyubibrahimi: That is what I assumed them to be. 

Because it's a list from the New Orleans District Attorney's office there should be events to extract from POST for New Orleans Police Department Officers and for New Orleans Sheriff's Department Officers. 

Is that incorrect?

Makefile
ayyubibrahimi: Ok. That makes sense. I will slow down. 

Because it was coming from a unique agency and because we have two brady lists from the New Orleans District Attorney's office (with potentially more coming), I was thinking of them as distinct. My bad.

match/new_orleans_da.py
ayyubibrahimi: We don't have a pprr for the New Orleans's Sheriff's office. We only have POST and per-new-orleans-so, which was extracted from cprr_new_orleans_so so I thought it was correct to match to POST. 

I now see why it is wrong for NOPD. Is this also wrong for NOSO?

clean/new_orleans_da_cprr_2021_brady.py
ayyubibrahimi: Resolved.

match/new_orleans_da.py
pckhoi: If you are going to lift roster data for an agency from POST then it's proper to do this. The point is to have consistent uid across tables. Since we are extracting roster data for NOSO from POST I guess it is okay.

pckhoi (approved):
LGTM

PR #31, opened 2021-07-01
Bug/st tammany so

ayyubibrahimi:


pckhoi:
Something strange is going on with this branch. I see the same changes that are already applied in a previous PR. And when I pull and compare with main I can't see anything new. @ayyubibrahimi any clue?

ayyubibrahimi:
We cleaned Brusly, madisonville, and IPM yesterday and pushed to enhancement/tangipahoa_so, but not st tammany which I cleaned today. Could that have something to do with it?

pckhoi:
Never mind then, there's no new changes on this branch regardless.

PR #43, opened 2021-07-06
Bug/events

ayyubibrahimi:


clean/grand_isle_pd.py
```diff
@@ -135,3 +139,144 @@ def clean():
     df.to_csv(data_file_path(
         'clean/pprr_grand_isle_pd_2021.csv'
     ), index=False)
+<<<<<<< HEAD
+=======
```
pckhoi, line 143: Again, unresolved conflicts.

match/grand_isle_pd.py
```diff
@@ -1,3 +1,7 @@
+<<<<<<< HEAD
+<<<<<<< HEAD
+=======
+>>>>>>> origin
```
pckhoi, line 4: unresolved conflict

clean/grand_isle_pd.py
ayyubibrahimi: I've been only treating unresolved conflicts only when github informs me that there is a conflict. I'll find a better way to be informed. Will fix these now.

match/grand_isle_pd.py
ayyubibrahimi: Resolved correctly this time, I believe. Please review.

clean/baton_rouge_csd_pprr.py
```diff
@@ -1,3 +1,4 @@
+from clean.scott_pd_pprr import clean_rank
```
pckhoi, line 1: This form of import is bad. What will happen if you want to update Scott PD? If you are not careful anything that import from `clean.scott_pd_pprr` will break and you have to fix them. As a general rule, things that can be shared must live under `lib` package. If you think this function is generic enough that it can be reused then please put it under `lib` package.

clean/kenner_pd.py
```diff
@@ -1,3 +1,4 @@
+from clean.scott_pd_pprr import clean_rank
```
pckhoi, line 1: Here too. Don't do this.

clean/baton_rouge_csd_pprr.py
ayyubibrahimi: DIdn't mean to do this. I've noticed that visual studio has been automatically calling functions that share the same name from different scripts. I missed this creeping into baton_rouge_csd_pprr.py and kenner_pd.py. Will resolved now.

pckhoi (approved):
LGTM

ayyubibrahimi:
Converting to draft as I'll use the events table to create references for rank_desc, dept_desc, and complaints.

PR #50, opened 2021-07-13
Bug/end of line

pckhoi:
Addressing following issues:
- Replace `CR LF` with `LF`
- Put correct dropbox links for St Tammany SO and LSP
- Canonicalize officer names for Tangipahoa SO CPRR
- Output personnel table for Tangipahoa SO


ayyubibrahimi (approved):
Approved.

PR #60, opened 2021-07-22
Enhancement/lafayette pd split action

pckhoi:
- `standardize_from_lookup_table` preserve original token order
- Split out to 1 allegation per row for Lafayette PD CPRR
- Split column `action` out of `disposition` for Lafayette PD CPRR


ayyubibrahimi (approved):
Approved

PR #64, opened 2021-08-06
Enhancement/baton rouge pd

ayyubibrahimi:


ayyubibrahimi:
@pckhoi - I stopped using standardize from look up table because because some charges would be divided by a semi-colon despite the code only referencing a slash. Happy to re-do it.

pckhoi:
Which tool to use is up to you. It is also possible to combine both approaches.

ayyubibrahimi:
I found combining both to be the best strategy to clean charges in this dataset, but this semi-colon kept appearing in place of a slash. I've been combining the two for the FDP datasets and it works without a problem.

pckhoi:
It sounds like you want to include one or more slashes in some output sequences? If it is so then you just need to include slash in the canonical sequence instead of splitting it up. For example:

```python
[
    # use this
    ["1-day suspension/loss of take home vehicle privileges for 30 days", ...],
    # instead of theses
    ["1-day suspension", ...],
    ["loss of take home vehicle privileges for 30 days",...],
]
```

If that's not what you meant then I'll be happy to take a look.

Now that I think about it more, ideally we want to use `standardize_from_lookup_table` for everything because it does not leave room for error, and the output is very clean. So it makes sense to get to the bottom of this.

ayyubibrahimi:
That's not what I mean. 

Sounds good. I'll reproduce it for your review.

ayyubibrahimi:
Also noticed that my flake8 isn't giving me warnings here. Pushing to draft.

ayyubibrahimi:
Nevermind. Look up table seems to be working now. Do you think I should add it to these scripts or move onto new_orleans_cprr_2020?

pckhoi:
I see 70 files have changed. This is very alarming. Do you know why? I suspect most files were changed due to LF being replaced with CRLF on Windows, but don't we already have EdtiorConfig to prevent that from happening?

ayyubibrahimi:
That is what happened. I reverted and only pushed the relevant changes. Thanks. That said, my editor wasn't giving me the appropriate errors for these two scripts. I plan on looking more into it tomorrow.

pckhoi:
@ayyubibrahimi is this PR ready for review now?

ayyubibrahimi:
Yes

clean/baton_rouge_so_cprr.py
```diff
@@ -21,18 +21,15 @@ def split_name(df):
 
 
 def clean_action(df):
-    df.loc[:, "action"] = df.action.str.lower().str.strip()\
+    df.action = df.action.str.lower().str.strip()\
```
pckhoi, line 24: Please use `.loc` in  assignment. This will probably generate a warning.

fuse/all.py
```diff
@@ -1,120 +1,120 @@
...
+    ensure_uid_unique(event_df, 'event_uid')
+    com_df = fuse_complaint()
+    ensure_uid_unique(com_df, 'complaint_uid')
+    com_df.loc[:, 'allegation_uid'] = ''
+    uof_df = fuse_use_of_force()
+    ensure_uid_unique(uof_df, 'uof_uid')
+    per_df.to_csv(data_file_path("fuse/personnel.csv"), index=False)
+    event_df.to_csv(data_file_path("fuse/event.csv"), index=False)
+    com_df.to_csv(data_file_path("fuse/complaint.csv"), index=False)
+    uof_df.to_csv(data_file_path('fuse/use_of_force.csv'), index=False)
```
pckhoi, line 120: This file is probably having the CRLF issue again? Please look into it.

fuse/ponchatoula_pd.py
```diff
@@ -1,49 +1,49 @@
...
+        post_event,
+    ]))
+    ensure_uid_unique(event_df, 'event_uid')
+    ensure_data_dir('fuse')
+    rearrange_personnel_columns(pprr).to_csv(data_file_path(
+        'fuse/per_ponchatoula_pd.csv'
+    ), index=False)
+    event_df.to_csv(data_file_path(
+        'fuse/event_ponchatoula_pd.csv'
+    ), index=False)
```
pckhoi, line 49: This file probably has CRLF issue too.

lib/standardize_test.py
```diff
@@ -1,26 +1,26 @@
...
+            ['sustained', 'sus', 'sus.'],
+            ['handling of evidence', 'evidence', 'handling'],
+            ['10 days suspension', 'sus 10', 'suspended 10 days'],
+            ['disobey direct order', 'disorbey']
+        ], True)
+        assert_series_equal(df2.disposition, pd.Series([
+            'handling of evidence; sustained; 10 days suspension',
+            'sustained; handling of evidence',
+            'disobey direct order; sustained; 10 days suspension'
+        ]), check_names=False)
```
pckhoi, line 26: And this file.

match/ponchatoula_pd.py
```diff
@@ -1,42 +1,42 @@
...
+
+
+if __name__ == '__main__':
+    pprr = pd.read_csv(data_file_path('clean/pprr_ponchatoula_pd_2010_2020.csv'))
+    post = pd.read_csv(data_file_path('clean/pprr_post_2020_11_06.csv'))
+    post_events = extract_post_events(pprr, post)
+    ensure_data_dir("match")
+    post_events.to_csv(data_file_path(
+        "match/post_event_ponchatoula_pd_2020.csv",
+    ), index=False)
```
pckhoi, line 42: And this file.

ayyubibrahimi:
Accidently permanently deleted your comments. I was using the delete as a check list for what errors to fix, without realizing I could not bring them back. My bad. 

Your comments were: 1) fix CRLF issue in match/ponchatoula and fuse/ponchatoula; 2) fix CRLF issue in fuse/all.py; 3) fix CRLF issue in standardize_test; 4) use loc in clean/baton_rouge_pd_cprr.py and clean/baton_rouge_so_cprr.py.

CRLF should no longer be an issue.

pckhoi:
@ayyubibrahimi I thought this should go without saying but deleting others's comments is a very bad idea. Now I have no way of knowing what feedback I gave. Please don't do that again.

lib/events_test.py
```diff
@@ -1,175 +1,175 @@
...
+                ['left', '05', 'def', '2000', '3', '28'],
+                ['left', '06', 'qwe', '2000', '4', '10'],
+                ['hire', '07', 'abc', '1990', '5', '4'],
+                ['hire', '08', 'abc', '1990', '4', '28'],
+                ['hire', '09', 'abc', '1990', '5', '24'],
+                ['hire', '10', 'def', '1990', '5', '4'],
+                ['hire', '11', 'def', '1990', '3', '28'],
+                ['hire', '12', 'qwe', '1990', '4', '10'],
+            ], columns=columns)
+        )
```
pckhoi, line 175: Looks like this file still has CRLF issue.

ayyubibrahimi:
I didn't know that deleting them in visual studio would delete them here. Very sorry!

ayyubibrahimi:
Please review. Thank you.

pckhoi (approved):
LGTM

PR #72, opened 2021-08-19
Enhancement/new orleans so

ayyubibrahimi:


ayyubibrahimi:
I didn't add this to wrglsync.yaml because I noticed that you didn't add it for new_orleans_so_cprr_19. Should that be the case?

pckhoi:
Yeah, as I said before I feel like putting the clean and match tables onto Wrgl serves no practical purpose for now.

clean/new_orleans_so_cprr.py
```diff
@@ -113,14 +273,222 @@ def split_investigating_supervisor(df):
...
+def clean_investigation_complete_date_20(df):
+    df.loc[:, 'investigation_complete_date'] = df.date_completed.fillna('')\
+        .str.replace(r'^(\d+)//(\d+)', r'\1/\2', regex=True)\
+        .str.replace('031/2/2020', '3/12/2020', regex=False)\
+        .str.replace('3/42021', '3/4/2021', regex=False)
+    return df.drop(columns='date_completed')
+
+
+def clean_action_20(df):
+    df.action = df.action.str.lower().str.strip()\
```
pckhoi, line 302: Please use `.loc` here. Maybe we can use an automated linter for these things.

ayyubibrahimi: I will look into it.

ayyubibrahimi:
Thank you for the reminder.

pckhoi:
let me know when this is ready for review.

fuse/new_orleans_so.py
```diff
@@ -10,28 +10,86 @@
...
+            'prefix': 'arrest',
+            'parse_date': True,
+            'keep': ['uid', 'agency', 'complaint_uid'],
+        },
+        events.OFFICER_SUSPENDED: {
+            'prefix': 'suspension',
+            'parse_date': True,
+            'keep': ['uid', 'agency', 'complaint_uid'],
+        },
+        events.OFFICER_RESIGNED: {
```
pckhoi, line 41: Should probably use `OFFICER_LEFT` instead. And you can add a `left_reason` column to the event table.

ayyubibrahimi: Ok.

ayyubibrahimi:
Please review @pckhoi

pckhoi:
For cases where people left on different dates, I guess they could just be separate events. Are there people left and then rehired and then left again on the same day though? I think the best we can do in those cases is explaining what happened in the `left_reason` column.

ayyubibrahimi:
I don't see any cases where people left and then were rehired and then left again on the same day. I doubt we'll have many instances of this, and if we do it's trash. 

I should make suspension, termination, and arrest distinct events then? Excepting the handful of cases where people were arrested and terminated on the same day

pckhoi:
Let me reword how I think about left event. If people left then they need to be rehired at some point for there to be another left event, regardless of whether we actually have data about when they are rehired or not. Any event that fits this semantic should be a left event. That's has always been the (perhaps implicit) definition for me.

We should definitely make arrest and suspension distinct events from termination. Regarding suspension, we could add suspension start and suspension end events.

ayyubibrahimi:
Sounds good.

ayyubibrahimi:
Please review. 

Only the last `officer_left` event is being extracted to the events table. For example, `officer_left `events with the prefix 'termination' is added to the events table while `officer_left` with the prefix 'resignation' is excluded. 

While reading through events.py I noticed the function `def discard_events_occur_more_than_once_every_30_days`. I think that it's probable we'll come across officers who were suspended more than once in 30 days.

pckhoi (approved):
LGTM

pckhoi:
@ayyubibrahimi that was probably set up to deal with repeated hire and/or left events. Perhaps we should deal with problematic datasets like that individually?

PR #75, opened 2021-09-03
Enhancement/new orleans pd

ayyubibrahimi:
Please review. 

I added an award_reccomended to the events table. What do you think?


match/new_orleans_pd.py
```diff
@@ -1,7 +1,7 @@
 from lib.date import combine_date_columns
 from lib.path import data_file_path, ensure_data_dir
 from datamatch import (
-    ThresholdMatcher, JaroWinklerSimilarity, DateSimilarity, ColumnsIndex
+    ThresholdMatcher, JaroWinklerSimilarity, DateSimilarity, ColumnsIndex, NoopIndex, Swap
```
pckhoi, line 4: `Swap` seems to be an extraneous import?

pckhoi (approved):
Looks good to me overall. The new event also looks good.

match/new_orleans_pd.py
ayyubibrahimi: Thanks.

PR #80, opened 2021-09-11
Enhancement/hammond pd

ayyubibrahimi:


clean/hammond_pd_cprr.py
```diff
@@ -148,7 +190,31 @@ def clean():
...
+        .pipe(gen_uid, ['first_name', 'last_name', 'agency'])\
+        .pipe(gen_uid, ['uid', 'charges', 'tracking_number', 'disposition'], 'complaint_uid')
+    return df
+
+
 if __name__ == '__main__':
-    df = clean()
+    df20 = clean_20()
+    df14 = clean_14()
     ensure_data_dir('clean')
```
pckhoi, line 218: FYI, we don't need to call `ensure_data_dir` ever again. Make took care of it.

pckhoi (approved):
LGTM

clean/hammond_pd_cprr.py
ayyubibrahimi: Got it. Thanks.

PR #99, opened 2021-09-21
Enhancement/universities

ayyubibrahimi:
Department pages for each university.


raw_datasets.json
```diff
@@ -179,5 +179,110 @@
...
+    "https://www.dropbox.com/s/f0m3ym4vvk66134/pprr_post_2020_11_06.csv?dl=1"
+  ],
+  "university_of_new_orleans_pd": [
+    "https://www.dropbox.com/s/f0m3ym4vvk66134/pprr_post_2020_11_06.csv?dl=1"
+  ],
+  "university_of_southwestern_louisiana_pd": [
+    "https://www.dropbox.com/s/f0m3ym4vvk66134/pprr_post_2020_11_06.csv?dl=1"
+  ],
+  "xavier_university_pd": [
+    "https://www.dropbox.com/s/f0m3ym4vvk66134/pprr_post_2020_11_06.csv?dl=1"
```
pckhoi, line 286: Duplicating the same file across multiple folders seems wasteful to me. Duplicating the same script over many files just changing the agency also seems wasteful. We don't have to stick to the implicit convention of one script/source file per agency if it is going to cause performance problems later. I would argue that the convention should be one clean script per one source file or a couple of highly related source files. It's not enough if we just clean the data, we also have to make sure the code remains easy to maintain and runs fast.

ayyubibrahimi: That makes sense. I guess that I was thinking ahead in that we'll eventually have distinct inputs for each university because I expect to receive roster data, eventually, and so I wanted to only use post as a place holder so that we can tag news articles to officers. But it makes a lot more sense to build out those entries as we receive data. I'll make the changes.

ayyubibrahimi:
Please review @pckhoi

pckhoi (approved):
Looks fine if you are confident there will be more data per university soon.

pckhoi:
But before you merge please keep in mind that if we want to change something about how we process the POST data or if we receive a more up-to-date POST data file, now you have to go make the changes in a dozen places. That's not a pretty sight. I still would not split the processing into multiple files. I think it's better to combine the processing of all universities to just one script for now. Nothing prevents you from producing multiple outputs from just one script file.

ayyubibrahimi:
Makes sense. I've consolidated the various university scripts into one.

pckhoi:
Please merge whenever you're ready.

PR #101, opened 2021-09-21
Enhancement/plaquemines so

ayyubibrahimi:


match/plaquemines_so.py
```diff
@@ -33,11 +34,34 @@ def match_post(cprr, post):
...
+    matches = matcher.get_index_pairs_within_thresholds(lower_bound=decision)
+    return extract_events_from_post(post, matches, 'Plaquemines SO')
+
+
 if __name__ == '__main__':
     cprr = pd.read_csv(data_file_path('clean/cprr_plaquemines_so_2019.csv'))
+    pprr = pd.read_csv(data_file_path('clean/pprr_plaquemines_so_2018.csv'))
     post = prepare_post_data()
-    cprr = match_post(cprr, post)
+    cprr = match_cprr_post(cprr, post)
```
pckhoi, line 61: Perhaps we should match CPRR against PPRR instead since the roster now comes from PPRR.

ayyubibrahimi: Yes. Thank you.

ayyubibrahimi:
Please review @pckhoi

pckhoi (approved):

PR #105, opened 2021-09-23
Bug/misc

ayyubibrahimi:


clean/louisiana_state_csc_lprr.py
```diff
@@ -9,7 +10,13 @@
 
 
 def standardize_appealed(df):
-    df.loc[:, "appealed"] = df.appealed.str.strip()
+    df.loc[:, "appealed"] = df.appealed.str.strip().fillna("")\
+        .str.replace(r"^(\w+) (\w+)", r'\1 - \2')\
```
pckhoi, line 14: Please specify the `regex` argument. Otherwise, the output will always contain warnings.

ayyubibrahimi: Thanks

pckhoi (approved):
LGTM

PR #110, opened 2021-09-29
Enhancement/new orleans so

ayyubibrahimi:


ayyubibrahimi:
Converting to draft until we receive further input from IPM or NOSO

lib/clean.py
```diff
@@ -237,13 +237,14 @@ def clean_races(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
             .str.replace(r'^w$', 'white', regex=True)\
             .str.replace(r'^h$', 'hispanic', regex=True)\
             .str.replace(r'^b$', 'black', regex=True)\
-            .str.replace(r'\bislande\b', 'islander', regex=True)
+            .str.replace(r'\bis?l?a?n?d?e?\b', 'islander', regex=True)\
```
pckhoi, line 240: this seems like it will match way too many strings. Even something like "ind" could match (where I think the intent is probably "native american"). It's better in this case to write out the variants in full so that this function won't make mistakes.

ayyubibrahimi: Agreed. I've thought about this a lot and haven't come to any conclusion because, for example,` (i|ind)` could refer to the Pacific Islands, Native America, or India. Thoughts?

pckhoi: Well, you can replace this pattern with strings that are clear in their intent. i.e. `islan`, `ilandr`, `islndr`, etc... It's better to be conservative here.

ayyubibrahimi: The citizen race string is `i`. Not sure If I'm missing your point.

pckhoi: I suppose you are thinking the intent for `i` is `islander`? For me the first thing comes to mind is `indian`. Are there other clues that tell you this is `islander`?

ayyubibrahimi: No, I just assumed islander because I've only ever seen someones race be listed as `islander `so far. Might have been a hasty judgement, but it's impossible to know because there are no names in this dataset.

pckhoi: Let's say that your assumption is correct. Even then you should make this transformation in the clean script, not in this function because it is shared among other clean scripts as well. In other places `i` might mean something else.

ayyubibrahimi: Sounds good.

pckhoi (approved):

PR #111, opened 2021-10-02
Enhancement/lake charles pd

ayyubibrahimi:


clean/lake_charles_pd_cprr.py
```diff
@@ -69,23 +141,23 @@ def split_rows_with_multiple_officers(df):
...
-def assign_empty_first_name_column(df):
+def assign_empty_first_name_column_20(df):
     df.loc[:, 'first_name'] = ''
     return df
 
 
-def review_first_names_from_post():
-    df = pd.read_csv(data_file_path('match/lake_charles_pd_cprr_2020_extracted_first_names.csv'))
+def review_first_names_from_post_20():
+    df = pd.read_csv(data_file_path('match/lake_charles_pd_cprr_2020_assign_first_name_from_post_pprr_2020_11_06.xlsx'))
```
pckhoi, line 154: Excel libraries in Python have the tendency to parse numerical and date/time values in strange ways which we have no control over. To ensure precise processing, we should only accept CSV as input.

pckhoi: Also, `make` might not be able to pick up input files other than CSVs.

clean/lake_charles_pd_cprr.py
```diff
@@ -120,30 +192,330 @@ def assign_agency(df):
...
+        parts = re.split(r"\s*(?:\/)\s*", s)
+        df = duplicate_row(df, idx + i, len(parts))
+        for j, name in enumerate(parts):
+            df.loc[idx + i + j, "officer_s_accused"] = name
+        i += len(parts) - 1
+    return df
+
+
+def review_first_names_from_post_19(df):
+    df = pd.read_csv(data_file_path('match/lake_charles_pd_cprr_2014_2019_assign_first_name_from_post_pprr_2020_11_06.xlsx'))
```
pckhoi, line 324: Same as above.

match/lake_charles_pd.py
```diff
@@ -51,12 +51,63 @@ def assign_uid_from_post(cprr, post):
...
     post = pd.read_csv(data_file_path('clean/pprr_post_2020_11_06.csv'))
     post = post[post.agency == 'lake charles pd']
     ensure_data_dir('match')
-    cprr = assign_first_name_from_post(cprr, post)
-    cprr = assign_uid_from_post(cprr, post)
-    cprr.to_csv(data_file_path(
+    cprr_20 = assign_first_name_from_post_20(cprr_20, post)
+    cprr_20 = assign_uid_from_post_20(cprr_20, post)
+    cprr_20 = assign_first_name_from_post_19(cprr_19, post)
+    cprr_20 = assign_uid_from_post_19(cprr_19, post)
```
pckhoi, line 109: Why not assign both first name and uid after a single matching? This is matching the same datasets twice, aka things run twice as long. Also, you are using different decisions and matching criteria for first name and uid. Doing this you might end up with people having the correct uid but incorrect first name or vice versa. I recommend just match the CPRR and POST once, then take both uid and first name from POST.

clean/lake_charles_pd_cprr.py
pckhoi: Never mind, this is not part of the processing, is it? Why not just put this in a notebook? A clean script importing things from the "match" directory seems backward to me.

pckhoi: Also, why do you need to view it in a data frame? I thought the Excel format makes it easy to review in Excel?

clean/lake_charles_pd_cprr.py
ayyubibrahimi: Makes much more sense to review in notebook. Thanks

match/lake_charles_pd.py
ayyubibrahimi: Makes sense. Thanks

pckhoi (approved):
LGTM

PR #115, opened 2021-10-02
Bug/ipm_nopd_cprr_allegations

ayyubibrahimi:
@pckhoi 

Do you remember why we decided to name the charges as `allegations` instead of `charges`?


pckhoi:
Sorry, I don't recall the reason. Let me ask the web guys to make sure.

ayyubibrahimi:
Also curious as to the thought behind the collapse of both `citizen_race` and `citizen_sex` into the one, `citizen` column``

pckhoi:
Regarding allegation vs charges, could you hold on until next week? I would like to confirm how these columns are being used and the web guys will not get back to me about it until then.

The citizen columns were collapsed because I think there were cases where there are multiple citizens per complaint. Will need to look through the script to make sure but that's the only reason I can think of.

ayyubibrahimi:
Sounds good

pckhoi (approved):
LGTM

PR #117, opened 2021-10-05
Enhancement/ouachita da

ayyubibrahimi:


match/ouachita_da.py
```diff
@@ -0,0 +1,42 @@
...
+    dfa = dfa.drop_duplicates(subset=['uid']).set_index('uid')
+
+    dfb = post[['uid', 'first_name', 'last_name', 'agency']]
+    dfb.loc[:, 'fc'] = dfb.first_name.fillna('').map(lambda x: x[:1])
+    dfb = dfb.drop_duplicates(subset=['uid']).set_index('uid')
+
+    matcher = ThresholdMatcher(ColumnsIndex(['fc']), {
+        'last_name': JaroWinklerSimilarity(),
+        'first_name': JaroWinklerSimilarity(),
+        'agency': JaroWinklerSimilarity(),
```
pckhoi, line 25: Are there lots of misspelling/typos for `agency`?

ayyubibrahimi: They're all missing

pckhoi: Sorry I'm missing the answer, are there typos for `agency`?

Just a reminder, these fields are all fuzzy-matched. While fuzzy matching is great for matching values with possible typos, it is a lot slower than simple `==` comparison. I find it quite odd that the `agency` field is also fuzzy-matched. Even if the matching is perfect, are you left with agency values with typos in them? And if there are no typos, you might want to put `agency` in `ColumnsIndex` instead which not only not making it slower, it actually makes matching faster.

PR #126, opened 2021-10-12
Miscellaneous bug fixes

pckhoi:
- Remove extraneous imports
- Eliminate empty sequence from lookup table in `clean_races`
- `write_deps.py` script output warning for files with no main block instead of failing.


pckhoi:
@ayyubibrahimi is there a reason why `new_orleans_booking.py` doesn't have a main block? Also FYI, going forward scripts without main block will be ignored by `make`.

ayyubibrahimi:
There is no reason for that at all. Not sure why that is the case. Will explore.

PR #130, opened 2021-10-13
Bug/new orleans pd

ayyubibrahimi:


clean/ipm_new_orleans_pd_cprr_allegations.py
```diff
@@ -207,6 +207,22 @@ def clean_investigation_complete_date(df):
     return df.drop(columns='completed_date')
 
 
+def drop_duplicate_complaint_uids(df):
+    return df.drop([9086, 15663])
```
pckhoi, line 211: This is very brittle. A change anywhere in the code could invalidate this index completely. A more robust way to drop duplications is to first sort the rows by disposition from more finalized to less finalized, then `drop_duplicates` with `keep='first'`. You also need to transform the `disposition` column to `CategoricalDtype` to make sorting possible. You can see how I did it in the `discard_allegations_with_same_description` function.

ayyubibrahimi: Amazing. Thank you.

PR #139, opened 2021-10-19
Enhancement/bossier city pd

ayyubibrahimi:


match/tangipahoa_da.py
```diff
@@ -0,0 +1,36 @@
...
+from lib.path import data_file_path
+import pandas as pd 
+
+
+def match_cprr_with_post(cprr, post):
+    dfa = cprr[['first_name', 'last_name', 'uid', 'agency']]
+    dfa.loc[:, 'fc'] = dfa.first_name.map(lambda x: x[:1])
+    dfa = dfa.drop_duplicates(subset=['uid']).set_index('uid')
+
+    dfb = post[['first_name', 'last_name', 'uid', 'agency']]
```
pckhoi, line 13: Don't you need to filter post data down to an agency first?

clean/tangipahoa_da.py
```diff
@@ -0,0 +1,33 @@
...
+def split_name(df):
+    names = df.name.str.lower().str.strip()\
+        .str.extract(r'(\w+) (\w+)')
+    df.loc[:, 'first_name'] = names[0]
+    df.loc[:, 'last_name'] = names[1]
+    return df
+
+def assign_agency(df):
+    df.loc[:, 'agency'] = ''
+    return df
```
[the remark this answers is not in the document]
ayyubibrahimi: No. This is not ready. Just created a draft pull request for enhancement/tangipahoa_da. I am waiting on a response from the Tangipahoa DA as to what agency these officers belong to as most of them belong to distinct agencies.

match/tangipahoa_da.py
ayyubibrahimi: Not yet, because most of them belong to distinct agencies, and, at the moment, I don't know what those agencies are. I was just exploring. I will filter by the agency once I have an agency for each officer. 

That said, if I'm unable to acquire the corresponding agency for each officer I think that I will stick with attempting to match to the entire post dataset.

PR #148, opened 2021-10-31
Enhancement/hammond pd

ayyubibrahimi:


clean/hammond_pd_cprr.py
```diff
@@ -163,6 +163,123 @@ def clean_action_09(df):
...
+
+def extract_and_clean_dispositions_08(df):
+    df.loc[:, 'dismiss'] = df.dismiss.str.lower().str.strip()\
+        .str.replace('x', 'dismissed', regex=False)
+
+    df.loc[:, 'disposition'] = df.founded_un.str.lower().str.strip()\
+        .str.replace(r'^u$', 'unfounded', regex=True)\
+        .str.replace('x', 'founded', regex=False)
+
+    df.loc[:, 'disposition'] = df.disposition.fillna('') + '' + df.dismiss.fillna('')
```
pckhoi, line 247: Perhaps you meant to put a single space instead of an empty string between these?

ayyubibrahimi: Fixed this. No separation is needed because there is no overlap, instead dismissed is a missing disposition.

pckhoi:
Please put the files in `fuse/all.py` and `data/.wrglsync.yaml`.

ayyubibrahimi:
The files are showing as already present in `fuse/all.py` and `data/.wrglsync.yaml` from when the older cprr tables were added.

pckhoi (approved):

PR #149, opened 2021-10-31
Bug/misc

ayyubibrahimi:


lib/clean.py
```diff
@@ -237,19 +237,15 @@ def clean_races(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
...
-            .str.replace(r'\bislande\b', 'islander', regex=True)\
-            .str.replace(r'^a$', 'asian', regex=True)
+            .str.replace(r'\bislande\b', 'islander', regex=True)
         df = standardize_from_lookup_table(df, col, [
             ['black', 'african american', 'black / african american', 'black or african american'],
             ['white'],
             ['hispanic', 'latino'],
             ['native american', 'american indian', 'american indian or alaskan native'],
-            ['asian / pacific islander', 'asian', 'native hawaiian or other pacific islander', 'islander'],
+            ['asian / pacific islander', 'asian', 'native hawaiian or other pacific islander'],
```
pckhoi, line 246: Are you sure that "islander" shouldn't be in this group? Doing this might break other scripts.

pckhoi:
Very nice typo corrections!

lib/clean.py
ayyubibrahimi: It should. Thanks.

pckhoi (approved):
LGTM

PR #160, opened 2021-11-14
enhancement/new_orleans_pd

ayyubibrahimi:
@pckhoi 

Hi Khoi,

Could I get your advice? I haven't been able to figure out how to get your function `split_rows_with_multiple_officers` to effectively work on this table. I'm currently working under the assumption that the issue lies in the number of rows contained in this table as the function continues to run after I wait 20 minutes for it to reach completion.

### DIFF
--- a/clean/new_orleans_pd_stop_and_search.py
+++ b/clean/new_orleans_pd_stop_and_search.py
@@ -1,4 +1,5 @@
 import pandas as pd
+from pandas.io.parsers import read_csv
 from lib.clean import clean_datetimes, float_to_int_str, standardize_desc_cols, clean_races
 from lib.path import data_file_path
 from lib.columns import clean_column_names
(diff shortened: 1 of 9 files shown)


clean/new_orleans_pd_stop_and_search.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def split_names(df):
+    return df
+
+
+def clean():
+    df = pd.read_csv(data_file_path('raw/ipm/new_orleans_pd_stop_and_search_2007_2021.csv'))\
+        .pipe(clean_column_names)\
+        .drop(columns=['subjectage'])\
```
pckhoi, line 191: The raw file seems to be the wrong name? Also when I changed the name it shows me this error: `KeyError: "['subjectage'] not found in axis"`. Any idea? Also curious how the file name can be wrong in the first place. Are you not using `scripts/rawfiles.sh` to get the raw file name perhaps?

ayyubibrahimi: Sorry. I added a version that I cleaned for some initial analysis to the `raw_datasets.json`. I just replaced it with the raw version.

pckhoi:
So the good news is that it doesn't look like an infinite loop or if there's a bug. There are just too many rows (194716), and the current method is only fast enough for several thousand rows or less. We could do some serious benchmarking to ascertain but I suspect it is slow due to the usage of Python's for loop. Nothing wrong with for loop, it's just mega-slow compared to vectorized operations by Pandas. I think it's time to cook up a new version of the duplicate row functionality.

ayyubibrahimi:
This solution seems to work pretty well. What do you think?

clean/new_orleans_pd_stop_and_search.py
```diff
@@ -0,0 +1,236 @@
...
+        .str.replace(r'(\w+)\/(\w+) ', r'\1.\2', regex=True)
+    return df.drop(columns=['officer1badgenumber', 'officer2badgenumber', 'officer1name', 'officer2name'])
+
+
+def split_rows_with_multiple_officers(df):
+    df = df.drop('officer_names_and_badges', axis=1).join(df['officer_names_and_badges']\
+        .str.split('/', expand=True)\
+        .stack()\
+        .reset_index(level=1, drop=True).rename('officer_names_and_badges'))\
+        .reset_index(drop=True)
```
pckhoi, line 170: Have you checked how many rows the final frame has? You might need to add `how='outer'` to `join` method. Also there's no need for backslash within brackets.

ayyubibrahimi: Thanks. The final frame has 830301 rows.

pckhoi (approved):

clean/new_orleans_pd_stop_and_search.py
pckhoi: Very nice change btw!

ayyubibrahimi: Awesome. Thanks. Converting to draft to complete matching and fuse.

ayyubibrahimi:
This is now ready for review. Thanks!

pckhoi (approved):
Is this still a draft?

ayyubibrahimi:
Ready for review. All good?

pckhoi:
Yes, feel free to merge now.

PR #164, opened 2021-11-16
Enhancement/consolidate columns info

pckhoi:
From now on, the only valid place to edit schema is the `data/datavalid.yml`.


ayyubibrahimi (approved):
Awesome.

PR #177, opened 2021-11-24
Fixing miscelaneous bugs Nov 24 2021

pckhoi:
- Rebuild fused file whenever content of `datavalid.yml` changes.
- (temporary fix) Drop rows with duplicated `stop_and_search_uid`.
- Drop unused columns from `datavalid.yml`.
- Replace `data/fuse/complaint.csv` with `data/fuse/allegation.csv`.


ayyubibrahimi (approved):
Awesome. Thanks.

PR #181, opened 2021-11-30
Fix datavalid version and deduplicate columns

pckhoi:


pckhoi:
@ayyubibrahimi this seems to fix it, remember to run `pip install -r requirements.txt` first.

ayyubibrahimi (approved):
Awesome.

PR #187, opened 2021-12-03
Add entity resolution to cross-agency matching

pckhoi:
- Introduce common name penalty to the matching
- Introduce entity resolution to keep `person_id` as stable as possible.


pckhoi:
@ayyubibrahimi please review when you have time

ayyubibrahimi (approved):
Looks awesome.

PR #191, opened 2021-12-03
enhancement/denham_springs_pd

ayyubibrahimi:


data/datavalid.yml
```diff
@@ -158,6 +158,9 @@ schemas:
       - name: investigator_uid
       - name: investigator_rank
       - name: shift_supervisor
+      - name: allegation_desc
+        description: >
+          text description of the allegation
```
pckhoi, line 163: @rajivsinclair  Should we rename this field to `summary` to keep it consistent with previous work à la CPDP?

ayyubibrahimi: I think that `allegation_summary` or `allegation_desc` is more appropriate as `summary` seems vague.

rajivsinclair: I think `allegation_desc` or `allegation_summary` is okay here.

pckhoi: that's fine then

pckhoi (approved):
LGTM

PR #207, opened 2021-12-09
analysis/cprr_nopd

ayyubibrahimi:
Hey @pckhoi - I wrote this with the intention of finding every instance in which a `tracking_number` was associated with multiple officers. In other words, to find out which, if any, officers frequently partner when engaging in misconduct. How does it look to you? I don't intend to push this code anytime soon. Just wanted your opinion.


match/allegations_clusters_nopd.py
```diff
@@ -0,0 +1,41 @@
...
+import numpy
+
+
+def cluster_officers_by_tracking_number(cprr):
+    df = cprr[['uid', 'first_name', 'last_name', 'tracking_number']]
+    df = df.drop_duplicates(subset='uid', keep='first').set_index('uid')
+    df.loc[:, 'tn'] = df.tracking_number.fillna('').map(lambda x: x[:])
+    matcher = ThresholdMatcher(ColumnsIndex('tn'), {
+        'tracking_number': StringSimilarity()
+    }, df, show_progress=True)
```
pckhoi, line 15: so it looks like `tn` column is almost exactly the same as `tracking_number`? Well, `ColumnsIndex` already ensures that rows with different `tn` do not get to match with each other so what is likely to happen is you only see exact matches (score is never below 1). Is that the intention? If it is then something like this is both shorter and run faster:
```python
df = df.set_index('tracking_number')
clusters = dict()
for tn, row in df.iterrows():
    clusters.setdefault(tn, list()).append(row)
# discard tracking numbers that involve less than 2 officers
clusters = {tn: l for tn, l in clusters.items() if len(l) > 1}

File: match/allegations_clusters_nopd.py [hunk 2]
```diff
@@ -0,0 +1,41 @@
...
+        dfa = pd.DataFrame(c)
+
+        dfa['clusters']= dfa.values.tolist()
+        dfa.loc[:, 'clusters'] = dfa.clusters.astype(str).str.lower().str.strip()\
+            .str.replace(r'(\, none)+', '', regex=True)
+        
+        numpy.sort(dfa.clusters, axis=-1, kind='quicksort')
+        
+        dfa = dfa[['clusters']]
+        dfa.to_excel(data_file_path('match/nopd_allegation_clusters_sorted.xlsx'))
```
pckhoi, line 34: Are you sure this entire logic should be repeated for every cluster? What is the intention here?

match/allegations_clusters_nopd.py
ayyubibrahimi: That is the intention. Thanks for the feedback!

match/allegations_clusters_nopd.py
ayyubibrahimi: By appending each row to an array and then normalizing the order of the arrays by sorting them, I hoped to be able to effectively check for duplicates which would allow me to deduce which officers were grouped to the same misconduct allegation in more than one instance.

match/allegations_clusters_nopd.py
pckhoi: Sorry, even my first attempt is the wrong way to do it ;). It could even be shorter and faster:
```python
grouped = cprr[['uid', 'first_name', 'last_name', 'tracking_number']]\
    .groupby("tracking_number").filter(lambda x: len(x) > 1)
```
I'm sure whatever else you want to do can be done with just "group by" operations. Can you read [this article](https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#filtration) and try to do everything with just operations discussed there?

ayyubibrahimi: Beautiful. I initially started with a "group by" operation as I've been doing all other analysis with them, but I didn't think to filter with a lambda function. Awesome. Thank you!

ayyubibrahimi:
@pckhoi - Please check out the new code I just pushed when you have the time. I don't think you'll love it. It's pretty long. 

It looks like there are 387 perfect duplicates. 

I still have more work to do because I need to find the imperfect duplicates, i.e., instances in which an additional officer is associated with a `tracking_number` that officers, who have a prior history, are also associated with.

pckhoi:
@ayyubibrahimi doesn't look like you have pushed the new code yet. Would you say that you are trying to clusterize complaints such that complaints in the same cluster must all share at least 2 uids perhaps? And then complaints that share more uids should score higher? If you look at the `entity_resolution` function in `match/cross_agency.py`, there is something similar occuring:

```python
def entity_resolution(old_person: pd.DataFrame, new_person: pd.DataFrame) -> pd.DataFrame:
    # both frames have 'person_id' which is the index and 'uids' which is uids separated by commas
    dfa = new_person.set_index('person_id', drop=True)
    # transform uids into set
    dfa.loc[:, 'uids'] = dfa.uids.str.split(pat=',')\
        .map(lambda x: set(x))

    dfb = old_person.set_index('person_id', drop=True)
    dfb.loc[:, 'uids'] = dfb.uids.str.split(pat=',')\
        .map(lambda x: set(x))

    # in new version of datamatch, you can specify a custom scorer with an arbitrary function
    def person_scorer(a: pd.Series, b: pd.Series) -> float:
        # here the score is 1 if canonical_uid is the same
        # but you can modify so that the score is 0 if a & b do not share
        # at least 2 officers
        if a.canonical_uid == b.canonical_uid:
            return 1
        # here you can see the score is higher if the percentage of shared
        # officers is higher
        x_len = len(a.uids & b.uids)
        return x_len * 2 / (len(a.uids) + len(b.uids))

    print('matching old and new person table...')
    matcher = ThresholdMatcher(
        # since uids are sets now, remember to use index_elements=True to
        # index with individual uid in each set. This mean that rows that share
        #  at least 1 uid will be able to match
        index=ColumnsIndex('uids', index_elements=True),
        scorer=person_scorer,
        # this matches 2 frames but the techniques should work with deduplication as well
        dfa=dfa,
        dfb=dfb,
        show_progress=True,
    )
    # decision is set this low so that everything except pairs that score 0 are output
    # to the excel file
    decision = 0.001
```

Sorry I veered back to using datamatch to solve it but I guess if you want to fuzzy match things and/or give a score to individual pair then datamatch is still the better route. I didn't understand what you were doing initially.

This is an example but you can get creative with it. The best way to start is to decide how you want to score the pairs, now that you can pass a function as scorer, the sky is the limit.

ayyubibrahimi:
Sorry, just pushed. 

Yes, that's about it. However, I'm not sure that a score should be amplified in importance because it is connected to a higher quantity of uids, necessarily. Instead, I hope to make clusters of uids based on the shared tracking_number, regardless of the number of uids associated with any given tracking_number. I've been approaching the problem from the point of view: what two uids are most frequently associated with the same tracking number?

pckhoi:
You can utilize `MultiIndex` to add another index based on tracking_number. That should prevent rows with different tracking_number from matching.

ayyubibrahimi:
@pckhoi - Pushed new code. Thoughts? 

It's pretty rudimentary as I ended up shying away from using datamatch after a bit more experimenting. From my understanding, the original code that I pushed using datamatch did the job as I'm only using datamatch to look for instances in which two, or more, officers share a `tracking_number`, and the groupby function performs faster.

The rest of the code is just wrangling to figure out which tuples of uids are associated with the same `tracking_number` more than once, as my goal is to figure out which groups of officers are frequently in association when accused of misconduct.

match/allegations_clusters_nopd.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def filter_out_non_duplicates(df):
+    #  creates a df in which a given tracking_number must appear more than once
+    df[['uid', 'tracking_number']].groupby("tracking_number").filter(lambda x: len(x) > 1)
+
+    #  if a given uid is associated with a given tracking number more than once, drop it
+    df['uids'] = list(zip(df.uid, df.tracking_number))
+    df = df[['uids']]
+    df = df.drop_duplicates(subset=['uids'])
```
pckhoi, line 16: These 3 lines could be replaced with
```python
df = df.drop_duplicates(subset=['uid', 'tracking_number'])
```

raw_datasets.json
```diff
@@ -67,7 +67,8 @@
     "https://www.dropbox.com/s/dnroduov6dnj5o3/job_code_description_list.csv?dl=1",
     "https://www.dropbox.com/s/if0g2l5muaccegt/new_orleans_so_charges_2015_2021.csv?dl=1",
     "https://www.dropbox.com/s/3nnvwsedqifwpoc/new_orleans_so_bookings_2015_2021.CSV?dl=1",
-    "https://www.dropbox.com/s/wajwv8qab6bsvwm/new_orleans_pd_stop_and_search_2007_2021.csv?dl=1"
+    "https://www.dropbox.com/s/wajwv8qab6bsvwm/new_orleans_pd_stop_and_search_2007_2021.csv?dl=1",
+    "https://www.dropbox.com/s/d59qrewhbbtl4yb/allegations_nopd_merged.csv?dl=1"
```
pckhoi, line 71: Is this a file that came from this processing pipeline?

match/allegations_clusters_nopd.py
```diff
@@ -0,0 +1,234 @@
...
+    for tuple in list_of_duplicates:
+        for row in uids:
+            if tuple in row and tuple not in list_of_duplicates:
+                print(row)
+    return df
+
+
+def clean():
+    df = pd.read_csv(data_file_path('raw/ipm/allegations_nopd_merged.csv'))\
+        .pipe(float_to_int_str, ['tracking_number', 'uid', 'uids'])\
```
pckhoi, line 225: `float_to_int_str` should only be used with decimal strings e.g. "2020". Which I think none of the columns qualify.

match/allegations_clusters_nopd.py
pckhoi: And it should be done before `groupby`.

match/allegations_clusters_nopd.py
```diff
@@ -0,0 +1,234 @@
...
+       'c75e6823cc3de53f0c4129af057acb4a, 40e190ec5d4a63c4e397e77f1410d2b5',
+       'd601272d7877017e5959c8dc67750ffa, d21f42a3196ab533161a1c446fe6f4d5',
+       '2b34d88c75e211c79055e5df7de66ecb, 71d3366908bf2f3d28d9e47cbbbe97a7',
+       'a50b15e88522ca16ca4901b5363730cf, d71bf1cd90cdf47820f7be40b7e0df7a',
+       '95f0da851956b555a435f511e6c76350, 19682fdd177ff6872cbef6f96a91c423',
+       '38655452c1a964184caf910e5be0c8ee, cb2c3497b94007fb1429c501da3c77b2',
+       '9aea0cd1efec57fa5b03d20f4d55b4be, 1a15535764f3158a07a75f9212319559',
+       '91f8792e8e1c08e379508f254f06a306, df9420dfb0af507791c67c401ca236f6',
+       '8174b1169b18edec85b8197d9e9d10b5, 88c9b8e47b0a813c2843e6d60e5eac57']
+    for tuple in list_of_duplicates:
```
pckhoi, line 216: This isn't a list of tuples, this is a list of str.

pckhoi:
@ayyubibrahimi If you are using datamatch for the right task, it is not slower. The only reason I suggested `groupby` at the start was that you seem to only want to match things absolutely. But then I see that you also want to match things "imperfectly". Which means that actually, datamatch will do the job better (and faster too).

match/allegations_clusters_nopd.py
ayyubibrahimi: Thank you.

raw_datasets.json
ayyubibrahimi: https://www.dropbox.com/s/d59qrewhbbtl4yb/allegations_nopd_merged.csv?dl=1 is df that is the product of a merge between two files from the processing pipeline. I won't be pushing this code to main.

match/allegations_clusters_nopd.py
ayyubibrahimi: Thanks.

match/allegations_clusters_nopd.py
ayyubibrahimi: Yes. Thank you.

ayyubibrahimi:
Makes sense. Thanks. I'm for sure doing something wrong. I'm going to dive back into it.

ayyubibrahimi:
This code seems to do the job.  What do you think? It's similar to the first code that I pushed, with the exception of the `uids `column. I don't think I properly understood your initial critique. 

    df.loc[:, 'uids'] = df.uid.str.cat(df.tracking_number, sep=', ')
    df = df[['uid', 'tracking_number', 'uids']].drop_duplicates(subset=['uids']).set_index('uids')
    df.loc[:, 'tn'] = df.tracking_number.fillna('').map(lambda x: x[:])
    matcher = ThresholdMatcher(ColumnsIndex('tn'), {
        'tracking_number': JaroWinklerSimilarity()
    }, df, show_progress=True)


Regarding the code that I pushed yesterday in which I solve the problem using groupby, do you see any issues besides what you commented above? Thanks!

pckhoi:
Let's hop on a LiveShare session whenever you have time. I think it will be easier to explain there.

ayyubibrahimi:
Cool. I'm getting off of a long bus now. How is your tomorrow morning/my tomorrow night or the day after?

pckhoi:
Sounds good to me!

ayyubibrahimi:
@pckhoi 

Pushed some new code. Integrated `pandas_dedupe`, which seems to work magically. Please review whenever you have the time.

pckhoi (commented):
The `dedupe` package (which powers the `pandas_dedupe` package) relies on machine learning which is fundamentally different from datamatch. And while it is a powerful approach, it does not necessarily fit the use case of this repo. We mostly deal with small datasets with zero training data, which is the exact kind of data that will benefit the least from machine learning. Not only does it not introduce any substantial improvement, but it also has many drawbacks. So I firmly oppose introducing `pandas_dedupe` into this repo for 5 reasons:
- Datamatch already do dedupe and unless absolutely necessary we should never introduce another package with similar purpose. Having multiple tools that do the same thing will very easily result in a messy codebase.
- Dedupe runs a lot slower. I will rewrite this script below, and minus the time spend saving data to Excel, datamatch runs in 1 second flat compared to 1-2 minutes run time for dedupe.
- The confidence score that dedupe produces is very opaque. You only know that a higher score mean a better match but nothing else. You can now specify your own scoring function for datamatch which uses simple calculation, thus the scores that come out are also very easy to understand. And hey, this problem that you are trying to solve is very simple in nature.

I rewrote the script below (please read the comments as well):
```python
import sys
sys.path.append('../')
import pandas as pd
from datamatch import ThresholdMatcher, ColumnsIndex
from lib.path import data_file_path


def create_clusters(df):
    # filter and group uids with tracking number
    df = df.loc[df.tracking_number.notna(), ['uid', 'tracking_number']]\
        .drop_duplicates()\
        .groupby('tracking_number').filter(lambda x: len(x) >= 2)
    df = pd.DataFrame(
        # we're storing set of uids under "uids" column because set operations
        # are very fast compared to string operations
        ((tn, set(sr.tolist())) for tn, sr in df.groupby('tracking_number')['uid']),
        columns=['tracking_number', 'uids']
    ).set_index('tracking_number', drop=True)
    print(df)
    #                                                               uids
    # tracking_number
    # 1999-0001-r      {fbe767d5226abd0ad6a5465d2f4ca1c8, 137328460cc...
    # 1999-0518-r      {77268e8c7996e93bed5dc4c1dc99fcc4, 77660eca405...
    # 2001-0519-c      {44da9db0f4c0705a37bacda5fba54219, 4225c320079...
    # 2002-0125-c      {3263d4b5b2790fe60b7a4d89ac76209e, fe0ea64005a...
    # 2003-0441-r      {3d6163a35d00431dc1a3044b207e471f, fe0ea64005a...
    # ...                                                            ...
    # 2019-0093-r      {c75e6823cc3de53f0c4129af057acb4a, da5d8837e26...
    # 2019-0095-n      {defb38a6722004c4dd6a7377946a3082, ca9c3b0f3ca...
    # 2019-0100-p      {3c72799f0a8079f0d080bbf2004dd860, c8283035ac0...
    # 2019-0109-p      {0717b9f92b35d1bca92c01a3a447ef31, 0d9a340272e...
    # 2019-0112-p      {29f585a7614321b970e865092bdb553c, b3cea4d1738...
    #
    # [2737 rows x 1 columns]

    # our custom scoring function
    def scorer(a: pd.Series, b: pd.Series) -> float:
        # x_len is the number of shared uid between 2 tracking numbers
        x_len = len(a.uids & b.uids)
        # if they only share 1 uid then they are not a match
        if x_len < 2:
            return 0
        # returns the ratio of shared uids
        # if the uids are exactly the same then the ratio will be 1
        # if no uid are shared then the ratio will be 0
        # but of course, we have already discarded such pairs in the above condition
        return x_len * 2 / (len(a.uids) + len(b.uids))

    matcher = ThresholdMatcher(
        # Splitting the "uids" column into distinct uid and index with those
        # distinct uid because index_elements is set to True.
        # As always, only the tracking_number that share a uid will be matched
        index=ColumnsIndex('uids', index_elements=True),
        # use the above scoring function
        scorer=scorer,
        dfa=df,
        show_progress=True
    )
    # decision is set to very low (but above 0)
    # to include pairs that score as low as 0.307692
    decision = 0.1
    matcher.save_clusters_to_excel(
        data_file_path("match/nopd_allegation_clusters.xlsx"), decision, decision
    )
    return matcher.get_clusters_within_threshold(decision)


if __name__ == '__main__':
    df = pd.read_csv(data_file_path('raw/ipm/allegations_nopd_merged.csv'))
    clusters = create_clusters(df)
    print(clusters)
    #                                                                                          uids
    # cluster_idx pair_idx sim_score row_key                                                       
    # 0           0        1.000000  2010-1039-i  {81ce213189ee816017229582c6d337ad, 009cbd42a2d...
    #                                2011-0565-c  {81ce213189ee816017229582c6d337ad, 009cbd42a2d...
    # 1           0        1.000000  2007-0510-c  {8e76e6643edc61dd8a6958a25a347c49, 40c2bcc1073...
    #                                2009-1223-n  {8e76e6643edc61dd8a6958a25a347c49, 40c2bcc1073...
    # 2           0        1.000000  2006-0195-n  {2d2cad8b9201f9d047ed71b5f54265f1, 7fc8478fc39...
    # ...                                                                                       ...
    # 345         0        0.307692  2018-0347-p  {a5b0b01a80ec49ecf72f1a270b8bfcbd, 12d38303dd2...
    # 346         0        0.307692  2013-0644-r  {ad896612811f86e9e6cd8e4bd40cadb9, 8a26726c831...
    #                                2014-0775-r  {dbbf11e24b26c9fd6fba88efa78c7581, 1142f363167...
    # 347         0        0.307692  2006-0449-c  {53b5a1a4f69b655d689a1ef9fdf2492f, 1a610dbe081...
    #                                2008-0165-n  {75fef3bbdf8abc3b622947f0a27d10d1, 18c1402fa57...

    # [1892 rows x 1 columns]
    clusters.to_csv(data_file_path('match/allegation_clusters_nopd_by_tracking_number.csv'), index=False)
```

As I said the scores that come out are very easy to understand. For example:
- 1.0 means all uids match
- 0.8 means 2-uids tn and 3-uids tn shares 2 uids (2 * 2 / (3 + 2) == 0.8)
- 0.667 means 3-uids tn and 3-uids tn shares 2 uids (2 * 2 / (3 + 3) == 0.667)

Before I think you mentioned that you don't care about the scores but in my opinion, scoring absolutely matters because in most cases, you want to be able to rank things in order of best to worst matches. And how are you going to do it without a score? Those who control the scoring scheme control the matching.

match/allegations_clusters_nopd.py
```diff
@@ -0,0 +1,64 @@
...
+def filter_out_non_duplicates(df):
+    #  if a given uid is associated with a given tracking number more than once, drop it
+    df = df.drop_duplicates(subset=['uid', 'tracking_number'])
+
+    #  creates a df in which a given tracking_number must appear more than once
+    df[['uid', 'tracking_number']].groupby("tracking_number").filter(lambda x: len(x) > 1)
+
+    #  remove rows with missing tracking_number
+    df.loc[:, 'tracking_number'] = df.tracking_number.fillna('')
+    return df[~((df.tracking_number == ''))]
```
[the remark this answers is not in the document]
ayyubibrahimi: Thanks!

match/allegations_clusters_nopd.py
```diff
@@ -0,0 +1,64 @@
...
+    df.columns = ['tracking_number', 'uid']
+
+    df.loc[:, 'uid'] = df.uid.str.lower().str.strip().fillna('')\
+        .str.replace(r'\n', ', ', regex=True)\
+        .str.replace(r'\, (\w+) ', '', regex=True)\
+        .str.replace(r'^(\w+)  +', '', regex=True)\
+        .str.replace(r'\,? name(.+)', '', regex=True)\
+        .str.replace(r'^(\w+)$', '', regex=True)\
+        .str.replace(r'  +', ', ', regex=True)
+    return df[~((df.uid == ''))]
```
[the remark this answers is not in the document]
ayyubibrahimi: Thanks!

ayyubibrahimi:
It's beautiful and works perfectly. Thank you for the guidance and the clear notes. 

Thank you, as well, for correcting me when I was very wrong about the importance of scores re the imperfect matches. Not sure what my rationale was at the time with that thought, but it was clearly foolish. 

I'm going to close this pull request now because we're still a few months away (launch date is currently, and tentatively, March 2020) from when further analysis is necessary re: producing insights that will be included in the LLEAD data insights report and LLEAD data insights page of llead.co. I'm sure I'll have more questions in the new year.

PR #210, opened 2021-12-13
enhancement/post

ayyubibrahimi:
- added a `source_agency` column to POST 
- added post_pprr to fuse 

@pckhoi - Tarak is using the personnel table to match officers with the minutes data that he is extracting. We discussed fusing post into the personnel table, because it currently only contains officers for whom we have either a corresponding complaints table or personnel table. Thoughts?


pckhoi:
Cool, my prime concern is what should we do with duplicated events and duplicated personnel rows that this will produce. If this should go into the personnel and event table then we should make sure to filter out the agencies that we are already processing. The question is, how can we do it in an automated way to minimize human error.

ayyubibrahimi:
I was hoping that you had an idea. Not sure. It seems like the safest thing is just to send Tarak the cleaned post table which he can use in addition to the personnel table that he's currently withing work?

pckhoi:
Well, I think it's worth trying to incorporate post into the main output if it has values for the LLEAD website too. And it's nicer for Tarak to process.

@rajivsinclair Do you think adding data for agencies that are only present in POST (a big roster but no complaint) would be a good thing for LLEAD?

ayyubibrahimi:
If we can easily remove the post rosters when we receive personnel data from the source agency, as it is collected, I'm fully in support.

rajivsinclair:
Yea, I think it's useful to include POST data. And we should merge in rosters as they become available.

pckhoi:
@ayyubibrahimi what do you think about adding data from POST right in the `fuse/all.py` script? Do it right after we fused everything else. That way we can filter out agencies we are already processing with something like 
```python
post = post.loc[~post.agency.isin(event.agency.unique())]
```

Also why do you need to add the `source_agency` column?

pckhoi:
Addition of `source_agency` sounds fine to me.

Do the brady lists come with dates of some sort. I guess it makes sense to integrate the brady lists into the event table so long as we can say something like "officer x was put on brady list y on date z". Are you thinking of events like "brady_list_addition" that references an officer and a brady list maybe?

pckhoi:
@ayyubibrahimi re POST integration, I remember there are many cases where the agency texts in POST data were transformed during match and fuse steps, and so might not match agency texts in other scripts. Perhaps you'll want to standardize agency text for POST during the clean step.

ayyubibrahimi:
Re brady list: I think the best we can do is something like "officer x was included on the brady list as of this date", i.e., the date on which we received the list. From my understanding, the DA's office doesn't do a good job of keeping track of when an officer is added to a brady list, as the list are compiled over many years and most likely in a haphazard manner. That said, the date on which we receive the list seems like a suitable substitution. Yes, I'm thinking of an event like "brady_list_addition" which references an officer and the corresponding brady list. 

Re POST: Great point. I'll work on converting the agency names to how they are originally represented in the POST table. Thanks.

pckhoi:
Can we do it the other way around? How about standardizing POST's agency according to the current agency format? I'm afraid if agency text changes then it will bring surprises to the web team.

ayyubibrahimi:
Makes perfect sense. I'll do that instead.

ayyubibrahimi:
@pckhoi - Just pushed some code. Is that what you had in mind?

clean/post_pprr.py
```diff
@@ -11,7 +11,11 @@
 
 def standardize_agency(df):
     df.loc[:, "agency"] = df.agency.str.strip()\
-        .str.lower().fillna("").str.replace(r"(\w)\.\s*(\w)\.", r"\1\2")
+        .str.lower().fillna("").str.replace(r"(\w)\.\s*(\w)\.", r"\1\2", regex=True)\
+        .str.replace(r'e\. baton rouge so', 'baton rouge so', regex=True)\
+        .str.replace(r'e\. jefferson levee pd', 'east jefferson levee pd', regex=True)\
+        .str.replace(r'^st ', 'st. ', regex=True)\
+        .str.replace(' parish', '', regex=False)
```
pckhoi, line 18: POST data should share the same agency text as the current output data which means:
- Agency should be in title case.
- Agency should end with either SO or PD (is that all the abbreviations?).
- We should remove all the agency assignments in all match scripts that read POST data. After all, the agency text should be correct at the clean stage.
- And finally, we should hunt down any deviation from this formula in the `fuse/all.py` script, make sure it raises an error if it finds an agency that is not found in POST data. Then we should update this `clean/post_pprr.py` script with the correct agency text.

fuse/all.py
```diff
@@ -170,3 +184,9 @@ def fuse_stop_and_search():
     allegation_df.to_csv(data_file_path("fuse/allegation.csv"), index=False)
     uof_df.to_csv(data_file_path('fuse/use_of_force.csv'), index=False)
     sas_df.to_csv(data_file_path('fuse/stop_and_search.csv'), index=False)
+    post_events_df = fuse_post_events()
+    post_events_df = post_events_df.loc[~post_events_df.agency.isin(event_df.agency.unique())]
```
pckhoi, line 188: As mentioned above, we should raise an error if there are any agency text in `event_df` that does not appear in POST, which would indicate that we need to make a change in the clean script. Unless there are actually agencies that do not appear in POST. Are there agencies like that @ayyubibrahimi ?

ayyubibrahimi: I'm on it. All official agencies do appear in POST.

clean/post_pprr.py
ayyubibrahimi: @pckhoi 
Pushed new code, please review when you have the time. 

- Abbreviations have been standardized as best as possible, i.e., PD, SO, and DA (for District Attorney's Offices in cases where we receive brady lists) have been standardized. There are some agencies that I'm still not sure how to abbreviate at the moment, like: "Abbeville City Marshall". I think that it's fine to leave them as is because we only intend to collect data from PDs, SOs, and DAs for the foreseeable future. 

- When I remove the agency assignments from match scripts that read POST data I find that officers are matched to officers from different agencies, i.e., "John Smith" from Baton Rouge SO is matched to "John Smith" from Lafayette PD instead of "John Smith" from Baton Rouge SO.  Thoughts? 

- Added a function that raises an error if an agency is in the event_df and not in the post_event_df

fuse/all.py
```diff
@@ -157,6 +157,15 @@ def fuse_stop_and_search():
     ])).sort_values(['agency', 'stop_and_search_uid'])
 
 
+def find_event_agency_if_missing_from_post(event_df, post_event_df):
+    missing_event_agency = event_df[~event_df['agency'].isin(post_event_df['agency'])]
+    missing_event_agency = missing_event_agency[['agency']].drop_duplicates()
+
+    for agency in missing_event_agency['agency']:
+        print('Error! Agency not in POST! \nAgency name: ', agency)
```
pckhoi, line 165: Better replace with this statement which stops the script dead on its track.
```python
raise ValueError('Agency not in POST!: "%s"' % agency)
```

pckhoi: Or rather replace the for loop with:
```python
if len(missing_event_agency['agency']) > 0:
    raise ValueError('Agency not in POST: %s' % missing_event_agency['agency'].tolist())
```

fuse/all.py
```diff
@@ -172,3 +181,15 @@ def fuse_stop_and_search():
     allegation_df.to_csv(data_file_path("fuse/allegation.csv"), index=False)
     uof_df.to_csv(data_file_path('fuse/use_of_force.csv'), index=False)
     sas_df.to_csv(data_file_path('fuse/stop_and_search.csv'), index=False)
+
+    post_event_df = pd.read_csv(data_file_path('fuse/events_post.csv'))
+    post_personnel_df = pd.read_csv(data_file_path('fuse/per_post.csv'))
+
+    missing_agency_df = find_event_agency_if_missing_from_post(event_df, post_event_df)
+    missing_agency_df.to_csv(data_file_path('fuse/event_agency_missing_from_post.csv'), index=False)
```
pckhoi, line 189: Since the script should ideally crash here, there's no need to save a CSV.

match/shreveport_pd.py
```diff
@@ -1,82 +1,82 @@
...
+    cb = pd.read_csv(data_file_path(
+        'clean/cprr_codebook_shreveport_pd.csv'
+    ))
+    cprr = match_cprr_codebook(cprr, cb)\
+        .pipe(gen_uid, ['agency', 'tracking_number', 'allegation'], 'complaint_uid')
+    cprr = match_cprr_post(cprr, post)
+    ensure_data_dir('match')
+    cprr.to_csv(data_file_path(
+        'match/cprr_shreveport_pd_2018_2019.csv'
+    ), index=False)
```
pckhoi, line 82: Is this file plagued by Windows newline issue again?

clean/post_pprr.py
pckhoi: Very nice! Re mismatching of "John Smith", I think it's probably a typo. I can't think of any valid reason.

pckhoi:
Very nice overall. Please make sure to merge changes from the main branch and run `make` again to ensure everything is in good shape. If `make` finishes successfully then it should mean that the POST data is integrated flawlessly.

ayyubibrahimi:
Thanks for the comments. I'll make those changes tomorrow. I ran make prior to pushing this code, and it worked so it looks like the POST data has been integrated. I'll be sure, of course, to run it again after I incorporate your notes.

ayyubibrahimi:
@pckhoi - Please review when you have the time. Thanks!

data/.wrglsync.yaml
```diff
@@ -0,0 +1,618 @@
...
+    primaryKey:
+      - person_id
+commit:
+  messageTemplate: |-
+    {{.GitCommit.Link}}
+    {{.GitCommit.Message}}
+  gitRemote:
+    host: github.com
+    dir: ppact/processing
+  gitRoot: ..
```
pckhoi, line 618: This file shouldn't be here. Have you merged the main branch into this branch yet?

match/shreveport_pd.py
```diff
@@ -1,82 +1,78 @@
...
+    ))
+    cb = pd.read_csv(data_file_path(
+        'clean/cprr_codebook_shreveport_pd.csv'
+    ))
+    cprr = match_cprr_codebook(cprr, cb)\
+        .pipe(gen_uid, ['agency', 'tracking_number', 'allegation'], 'complaint_uid')
+    cprr = match_cprr_post(cprr, post)
+    ensure_data_dir('match')
+    cprr.to_csv(data_file_path(
+        'match/cprr_shreveport_pd_2018_2019.csv'), index=False)
```
pckhoi, line 78: Again, this file seems to be plagued with Windows newline sequence.

match/shreveport_pd.py
ayyubibrahimi: Apparently. No idea why because I had set a default setting. I've gone ahead and standardized the lines formation for the `to_csv` functions in the match directory, to match your formatting from the match/baton_rouge_pd.py script. What do you think?

match/shreveport_pd.py
ayyubibrahimi: Apparently. No idea why because I had set a default setting. I've gone ahead and standardized the lines formation for the to_csv functions in the match directory, to match your formatting from the match/baton_rouge_pd.py script. What do you think?

data/.wrglsync.yaml
ayyubibrahimi: I have merged main into this branch. No idea why its still here. This file is no longer present in the other branches that I'm working from where I've merged main, i.e., the `allegations/houma_pd_2017_2018` branch that you approved yesterday.

ayyubibrahimi: I just merged the `allegations/houma_pd_2017_2018` branch with main, which itself correctly merged with main prior to pushing. After pushing `allegations/houma_pd_2017_2018`  to main, I again, as you'll see below, merged this branch with main, however, data/.wrgylsync.yaml is still present.

ayyubibrahimi: I figured it out because I just made the same mistake in the `enhancement/allegations_acadia_so branch`. 

Instead of proceeding with the `data/.wrgylsync.yaml` as deleted by you, I re-added the file and then merged. As you'll see in that branch, I deleted the  `data/.wrgylsync.yaml` file and added all relevant data to `wrgl/config.yaml `. 

In this branch, I've also just removed`data/.wrgylsync.yaml` and added post to `wrgl/config.yaml`

match/shreveport_pd.py
pckhoi: The line format isn't too important for me as long as they pass flake8 ;)

So you have no idea why this script has Windows's `\r\n`? Anyway, just change the newlines back to `\n` for now.

data/.wrglsync.yaml
pckhoi: It appears alright now :)

match/shreveport_pd.py
ayyubibrahimi: I have an inkling of an idea. 

I seldom receive a message that VS is reverting my lines to` \r\n ` and so I manually revert them back to `\n`. Where my understanding becomes fuzzy is when I try to figure out why it happens, infrequently at that, especially considering that I have changed my default settings. Will continue to figure out why this happens.  

I corrected the newline for `match/shreveport_pd`. Additionally, in the interest of standardization, I reformatted the lines in each match script to resemble the format that you originally used in `baton_rouge_pd`, which you'll see in the commit "standardize to_csv lines in match directory". If this was unnecessary, I'm happy to revert the changes.

data/.wrglsync.yaml
ayyubibrahimi: Awesome.

match/shreveport_pd.py
pckhoi: IMO, we should never spend time manually changing the code styling but rather leave it to auto-formatter like [black](https://dev.to/adamlombard/how-to-use-the-black-python-code-formatter-in-vscode-3lo0). I will open a PR for this so you can take a look.

ayyubibrahimi: Makes sense. That'd be awesome. Thanks.

ayyubibrahimi:
@pckhoi - Everything seems to be in order with this branch. I'm going to review again tomorrow, before asking for you to officially review. Have a great day.

pckhoi:
Good night! Thanks for your hard work!

ayyubibrahimi:
@pckhoi  - Confirmed. Please review. Thanks!

fuse/abbeville_pd.py
```diff
@@ -50,7 +45,7 @@ def fuse_events(cprr, post):
 
 if __name__ == "__main__":
     cprr = pd.read_csv(data_file_path("match/cprr_abbeville_pd_2019_2021.csv"))
-    post = prepare_post_data()
+    post = pd.read_csv(data_file_path("clean/pprr_post_2020_11_06.csv"))
```
pckhoi, line 48: Don't you need to filter by agency here?

ayyubibrahimi: 1000%. Thanks.

ayyubibrahimi: 1000%. Thanks.

fuse/baton_rouge_so.py
```diff
@@ -71,7 +64,7 @@ def fuse_events(cprr_18, cprr_21, post):
 if __name__ == "__main__":
     cprr_18 = pd.read_csv(data_file_path("match/cprr_baton_rouge_so_2018.csv"))
     cprr_20 = pd.read_csv(data_file_path("match/cprr_baton_rouge_so_2016_2020.csv"))
-    post = prepare_post_data()
+    post = pd.read_csv(data_file_path("clean/pprr_post_2020_11_06.csv"))
```
pckhoi, line 67: Here as well, if we don't filter POST by agency in every fuse script then event tables of individual agencies will include all post events, which is a lot of redundancy. You need to not only filter by agency in individual fuse script, but also do it in such a way that it's not possible for there to be 2 different agencies in the output because `fuse/all.py` script will not be able to detect that kind of error. I would change it like this:
```python
agency = cprr_20.agency[0]
post = pd.read_csv(data_file_path("clean/pprr_post_2020_11_06.csv"))
post = post.loc[post.agency == agency]
```

fuse/greenwood_pd.py
```diff
@@ -59,8 +44,8 @@ def fuse_events(cprr, post_pprr):
 
 
 if __name__ == "__main__":
-    cprr = prepare_cprr()
-    post = prepare_post()
+    cprr = pd.read_csv(data_file_path('clean/cprr_greenwood_pd_2015_2020.csv'))
```
pckhoi, line 47: Does this cprr have the correct agency text?

fuse/levee_pd.py
```diff
@@ -62,18 +62,8 @@ def fuse_events(cprr, post):
 if __name__ == "__main__":
     cprr = pd.read_csv(data_file_path("match/cprr_levee_pd.csv"))
     post = pd.read_csv(data_file_path("clean/pprr_post_2020_11_06.csv"))
-    post = post.loc[
-        (post.agency == "e. jefferson levee pd") | (post.agency == "orleans levee pd")
-    ]
```
pckhoi, line 67: I think you still need to filter here.

match/bossier_city_pd.py
```diff
@@ -69,9 +69,6 @@ def match_cprr_with_pprr(cprr, pprr):
     cprr = pd.read_csv(data_file_path("clean/cprr_bossier_city_pd_2020.csv"))
     pprr = pd.read_csv(data_file_path("clean/pprr_bossier_city_pd_2000_2019.csv"))
     post = pd.read_csv(data_file_path("clean/pprr_post_2020_11_06.csv"))
-    post = post[post.agency == "bossier city pd"]
```
pckhoi, line 72: Why do you not filter POST even in match script?

match/greenwood_pd.py
```diff
@@ -1,37 +1,51 @@
...
+        cprr.loc[cprr.uid.notna(), ["uid", "first_name", "last_name"]]
+        .drop_duplicates(subset=["uid"])
+        .set_index("uid", drop=True)
     )
-    dfa = gw_cprr[["first_name", "last_name", "mid"]].drop_duplicates().set_index("mid")
-    dfb = post_pprr[["last_name", "first_name"]].drop_duplicates()
+    dfa.loc[:, "fc"] = dfa.first_name.fillna("").map(lambda x: x[:1])
+
+    dfb = (
+        post[["uid", "first_name", "last_name"]]
```
pckhoi, line 18: Need to filter only POST rows for Greenwood PD here.

pckhoi:
@ayyubibrahimi please put back in filter clauses for POST in match and fuse scripts. This look like it will produce a lot of redundancy and long run time. Also please don't push any data yet. When you're finish making the changes, I want to check out this branch and review the data.

ayyubibrahimi:
I'm on it.

clean/post_pprr.py
ayyubibrahimi: I completely misunderstood what you meant here: "We should remove all the agency assignments in all match scripts that read POST data. After all, the agency text should be correct at the clean stage." Sorry! Corrections are in process.

fuse/greenwood_pd.py
ayyubibrahimi: Corrected.

fuse/baton_rouge_so.py
ayyubibrahimi: Thank you.

fuse/levee_pd.py
ayyubibrahimi: The output is correct using 

```
agency = cprr.agency[0]
post = pd.read_csv(data_file_path("clean/pprr_post_2020_11_06.csv"))
post = post.loc[post.agency == agency]
```

match/bossier_city_pd.py
ayyubibrahimi: Corrected.

match/greenwood_pd.py
ayyubibrahimi: I think that this is addressed.

ayyubibrahimi:
@pckhoi  - This branch is ready to be checked out for when you have the time to review.

pckhoi:
@ayyubibrahimi No problem. And thank you for your patience. I'll review tomorrow. :)

pckhoi:
@ayyubibrahimi it seems you forgot to push the latest code changes for this branch.

ayyubibrahimi:
@pckhoi - I somehow missed this notificaiton. Sorry. Please review.

fuse/cameron_so.py
```diff
@@ -40,7 +35,9 @@ def fuse_events(post):
 
 if __name__ == "__main__":
     cprr = pd.read_csv(data_file_path("clean/cprr_cameron_so_2020.csv"))
-    post = prepare_post_data()
+    agency = cprr.agency[0]
+    post = pd.read_csv(data_file_path("clean/pprr_post_2020_11_06.csv"))
+    post = post.loc[post.agency == agency]
```
pckhoi, line 40: This is fine but what would happen if there is a typo in the agency? We should raise an error if there's a mismatch in agency text also:
```python
if len(post) == 0:
    raise ValueError("agency not found", agency)
```

ayyubibrahimi: Yes. I'll make those changes now.

pckhoi: and to avoid repeating code we could just put all of this in a function with name like `lib.post.load_for_agency` or something:
```python
def load_for_agency(agency):
    post = pd.read_csv(data_file_path("clean/pprr_post_2020_11_06.csv"))
    post = post.loc[post.agency == agency]
    if len(post) == 0:
        raise ValueError("agency not found", agency)
    return post
```
Then you can call it like this:
```python
from lib import post
...
post_data = post.load_for_agency(agency)
```

ayyubibrahimi: Thanks

ayyubibrahimi:
@pckhoi - No hitch in the running of make. Please review

pckhoi (approved):
Couldn't find any problem.

ayyubibrahimi:
Cool. I'll take the steps to push to WrglHub tomorrow.

PR #211, opened 2021-12-13
Enhancement/pprr_jefferson_so

ayyubibrahimi:


clean/jefferson_so_pprr.py
```diff
@@ -0,0 +1,158 @@
...
+                     r' ?grants ?| ?budget ?| ?finance?i?a?l? ?| ?tax (collector|supervisor) ?|'
+                     r' ?public assisgments ?| ?youth programs ?| ?community relations ?|'
+                     r' ?lab services ?|\;? ?technical services| ?prisoner ?|'
+                     r' ?asset forfeure ?| ?strategic engagement ?| ?dlr ?|'
+                     r' ?tech services ?| ?sib ?| ?p\.o\.c\. ?| ?school offense ?|'
+                     r' ?accounts payable ?| ?range ?| ?light shop ?| ?public assignments ?|'
+                     r' ?help desk tech ?| ?print shop ?| ?mang services ?|'
+                     r' ?terminal agency ?| ?handgun perms ?| ?insurance ?| ?vice ?|'
+                     r' ?photo lab ?| ?fingerprint ?| ?polygraph ?| ?evidence ?|'
+                     r' ?digital forensic ?| ?dna ?| ?patrol ?)')
```
pckhoi, line 35: Be careful of very long regex. It tends to be slow. A lookup table will almost certainly run faster in this case. It will also minimize the corpus so that whatever additional regex that you have to use will also be simpler.

clean/jefferson_so_pprr.py
```diff
@@ -0,0 +1,158 @@
...
+        .str.replace(r'^serge$', 'sergeant', regex=True)\
+        .str.replace(r'^detective detective$', 'detective', regex=True)\
+        .str.replace(r'^sergeant patrol sergeant$', 'patrol sergeant', regex=True)\
+        .str.replace(r'\bpatroi\b', 'patrol', regex=True)\
+        .str.replace(r'\:$', '', regex=True)\
+        .str.replace('commanderssary', 'commander', regex=False)\
+        .str.replace(r' ?\-$', '', regex=True)\
+        .str.replace('coordinato', 'coordinator', regex=False)\
+        .str.replace(r'^repairman repairman$', 'repairman', regex=True)\
+        .str.replace('systmanager', 'systems manager', regex=False)
```
pckhoi, line 137: There's a very good use case for lookup table right here.

clean/jefferson_so_pprr.py
ayyubibrahimi: Thanks.

clean/jefferson_so_pprr.py
ayyubibrahimi: Sounds good. I'll make the change.

ayyubibrahimi:
@pckhoi  - I implemented your suggested edits. Please check them out when you have the time. Thanks!

pckhoi (approved):
LGTM

PR #212, opened 2021-12-16
Replace wrglsync functionality

pckhoi:


pckhoi:
@ayyubibrahimi I think you have the credentials to push already. So try to follow the README and let me know how it goes. Note that repositories that haven't already existed will be skipped during push/pull. In those cases, you can commit manually on wrglhub, or if you want something more automated please let me know.

ayyubibrahimi (approved):
Sounds good. I'll let you know how it goes once I give it a go.

PR #213, opened 2021-12-16
enhancement/allegations_acadia_so

ayyubibrahimi:


data/.wrglsync.yaml
```diff
@@ -576,6 +584,10 @@ repos:
     fileName: fuse/event_terrebonne_so.csv
     primaryKey:
       - event_uid
+  event-acadia-so:
+    fileName: fuse/event_acadia_so.csv
+    primaryKey:
+      - event_uid
```
pckhoi, line 590: Please merge the `main` branch into this branch and modify the `.wrgl/config.yaml` file instead.

ayyubibrahimi: Got it. Thanks.

ayyubibrahimi:
@pckhoi - This is ready for review whenever you have the availability. Thanks!

clean/acadia_so_cprr.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def clean():
+    df = pd.read_csv(data_file_path('raw/acadia_so/cprr_acadia_so_2018_2021.csv'))\
+        .pipe(clean_column_names)\
+        .rename(columns={
+            'date': 'receive_date'
+        })\
+        .pipe(clean_dates, ['receive_date'])\
+        .pipe(clean_names)\
```
pckhoi, line 39: Are the names in lower case and with punctuations removed? Even if they are I think it's still good practice to always use the `lib.clean.clean_names` function on the name columns no matter what. That way the names during matching are always uniform.

clean/acadia_so_cprr.py
```diff
@@ -0,0 +1,52 @@
...
+        .pipe(clean_allegation)\
+        .pipe(clean_action)\
+        .pipe(set_values, {
+            'agency': 'acadia parish so'
```
pckhoi, line 43: Shouldn't the agency name be in title case? So maybe `Acadia Parish SO` instead?

clean/acadia_so_cprr.py
ayyubibrahimi: Good call. Thank you.

clean/acadia_so_cprr.py
ayyubibrahimi: Yes. I should have held off on branch this until we've standardized agency names re: the POST branch. Thanks.

ayyubibrahimi:
@pckhoi - Please review when you have the time. Thanks.

pckhoi (approved):
LGTM

PR #215, opened 2021-12-19
allegations/houma_pd_2017_2018

ayyubibrahimi:


data/datavalid.yml
```diff
@@ -67,7 +67,10 @@ schemas:
           accused officer's unique identifier. This references the `uid` column in personnel.csv
       - name: tracking_number
         description: >
-          complaint tracking number from the agency the data originate from
+          internal affairs complaint tracking number from the agency the data originate from
+      - name: case_number
+        description: >
+          administrative complaint tracking number from the agency the data originate from
```
pckhoi, line 73: How is this distinct from `tracking_number`?

ayyubibrahimi: A `case_number`, in this specific table, is how the administrative department classifies investigations.

Thus far, we have been primarily receiving `tracking_numbers` produced by the internal affairs department. In most cases, we are receiving data directly from the internal affairs department, as they should be responsible for conducting investigations, however, in the case of Houma, the administrative department seems to have some responsibility for record keeping as well. Likely because they are a small department, therefore, I expect to see similar `case_number` fields in the future as we expand data collection to the smaller parishes/agencies. From my limited understanding of smaller agencies, so far, investigative responsibilities spill over into both the administrative departments and internal affairs departments, therefore, I expect that we'll continue to receive multiple tracking classifications.

pckhoi (approved):
LGTM

PR #217, opened 2021-12-21
Auto-format code with black

pckhoi:
- Add "black" as code formatter.
- Configure vscode to format code when focus shift.

@ayyubibrahimi all you need to do for this to work is to install black:
```
pip install black
```
Also please take care when merging a blacked branch and a non-blacked branch. I recommend running black on the non-blacked branch first:
```
black .
```
Then commit the changes, then try to merge.


ayyubibrahimi (approved):
Awesome. THanks.

PR #221, opened 2021-12-26
extraction/spacy

ayyubibrahimi:
Notes from yesterday: 

**Training**
Preprocessing of text files post OCR.: In this case, ~25% were used for training and ~25% were used for testing. Training data is uploaded to Doccano for supervised training with the use of custom created labels.

Atm, there are only 191 appeal hearing transcripts for NOPD, however, these documents are constantly produced and uploaded to their website and so I anticipate a constant intake of new data to process.

Additionally, I expect that other agencies produce similar appeal hearing transcripts, therefore, I hope to expand the model beyond NOPD data.

**Output**
Only 21/31 of the testing files (appeal hearing transcripts) were able to extract an accused_name, a prerequisite to generating a UID, using the current Spacy model. I'm pretty sure that I know why. I think that I overfit the training model with `accused_name` labels. Atm, I think that it should be an easy fix for the next iteration of producing training data.

As you can tell, I'm still very much in the beginning stages of creating an efficient pipeline, however, because I'm now at the stage where I have successfully been able to match the `uid` generated from the spacy model with an officer's corresponding `uid` in the personnel table, with my own rudimentary experiment, I thought that it would be a great time for you to review.

16/21 of the uid were successfully matched, with a score of 1, to a uid from the personnel table.

**Notes from my experience with integrating my local pipeline into LLEAD** 

1) I'm unable to download  files, from raw_datasets.json using my dropbox access key which exists in a `.txt` file.

2) Nor have I been able to download the `.jsonl` files, from raw_datasets.json, which contains the training data from Doccano. And  I've yet to figure out why. Therefore, all pdfs and jsonl files were manually copy and pasted into their respective directorys with the sole purpose/hope that we can create a pipeline that will allow for both txt and jsonl files to be downloaded. 

3) The biggest roadblock, as I currently understand, is that my pipeline has no integration whatsoever with the makefile responsible for producing an efficient dependency pipeline. 

@pckhoi  

Very appreciative of any and all feedback. Merry Xmas,


extract/db_download_files.py
```diff
@@ -0,0 +1,27 @@
+import dropbox
+from lib.get_dropbox import get_files, get_folders
+from lib.path import data_file_path
+
+
+with open(data_file_path("drop_box/token.txt", "r")) as f:
+    access_token = f.read()
```
pckhoi, line 7: The best practice to deal with access tokens and secrets of any kind is to store them within `.env` file as environment variables. And then load them during script execution using package `dotenv`.

extract/db_download_files.py
```diff
@@ -0,0 +1,27 @@
...
+print(folders)
+
+# # folder_id for transcripts folder within the directory above
+folder_id = "id:8ceKnrnmgi0AAAAAAAA5bA"
+
+# download directory
+download_dir = data_file_path("raw/new_orleans_pd/appeal_hearing_pdfs/")
+
+# # print('Obtaining list of files in target directory...')
+get_files(dbx, folder_id, download_dir)
```
pckhoi, line 27: Best practice is to put every line from line 6 and below (basically everything excepts imports) in the main block:
```python
if __name__ == '__main__':
    ...
```
That way these statements are only executed when you call the script directly, not when you merely import them. But even if you never import a script you should still use the main block as it communicates your intention clearer.

extract/ocr_db_data.py
```diff
@@ -0,0 +1,61 @@
...
+pyocr.tesseract.TESSERACT_CMD = r"C:/Program Files/Tesseract-OCR/tesseract.exe"
+
+
+dir_path = os.chdir(data_file_path("raw/new_orleans_csc/appeal_hearing_pdfs"))
+for file in os.listdir(dir_path):
+    if file.endswith(".pdf"):
+        tools = pyocr.get_available_tools()
+        if len(tools) == 0:
+            print("No OCR tool found!")
+            sys.exit(1)
```
pckhoi, line 23: Rather than print and then exit, just raise an error instead. Just use `Exception` in this case.

extract/spacy_model.py
```diff
@@ -0,0 +1,101 @@
...
+import spacy
+import pandas as pd
+import os
+from lib.path import data_file_path
+
+# #  import labeled data that has been exported from doccano
+labeled_data = []
+
+with open(
+    data_file_path("raw/new_orleans_csc/appeal_hearing_pdfs/training_data/ayyub.jsonl")
```
pckhoi, line 15: What tool did you use to create this file?

pckhoi: This comes from Doccano right?

pckhoi:
`raw_datasets.json` only deals with CSVs right now. I'm looking at extracting the dependency management and code organization into an independent framework, therefore I think we should not modify the current code base to accommodate OCR and NER. Let's wait until the new framework is finished and then try to integrate after that. In the meantime, just keep updating and improving this branch, and let me know if there are new changes that I should look at.

I have not used Spacy and Doccano before so I will need some time to get up to speed. The new framework will need to take into account OCR and NER tasks as well.

extract/db_download_files.py
ayyubibrahimi: Thanks.

extract/db_download_files.py
ayyubibrahimi: Got it.

extract/ocr_db_data.py
ayyubibrahimi: Smart. Thanks.

extract/spacy_model.py
ayyubibrahimi: Yes, this was jsonl is the output that I downloaded from Doccano after labeling my training set.

ayyubibrahimi:
Sounds perfect. Thanks so much for the quick review. I'll integrate your feedback and continue to work on overall efficiency. Merry Xmas!

raw_datasets.json
```diff
@@ -255,5 +256,11 @@
   ],
   "terrebonne_so": [
     "https://www.dropbox.com/s/7bewi8vdhl8tzu9/terrebonne_so_cprr_2019_2021_byhand.csv?dl=1"
+  ],
+  "drop_box": [
+    "https://www.dropbox.com/s/rhtao3b4liti8sw/token.csv?dl=1"
+  ],
```
pckhoi, line 262: I just noticed but is that your dropbox access token which is publicly shared on the internet?

ayyubibrahimi: That is correct. Just removed that. Thanks.

pckhoi: You should remove that file from Dropbox too just to be safe.

pckhoi:
@ayyubibrahimi what is the motivation for using Tesseract for OCR instead of Azure FR?

ayyubibrahimi:
No legitimate motivation. I started building this pipeline before we integrated Azure into our workflow. I just happened to choose Tesseract because I had used it on a prior project.

extract/match_spacy_w_wrgl_uid.py
```diff
@@ -0,0 +1,71 @@
...
+sys.path.append("../")
+import pandas as pd
+import os
+from datamatch import JaroWinklerSimilarity, ThresholdMatcher, ColumnsIndex
+from lib.path import data_file_path
+from lib.clean import clean_names
+import glob
+
+
+#  match appeal_uids with event_uids
```
ayyubibrahimi, line 12: @pckhoi the matching runs very slowly. Is it due to what you were referencing recently regarding the added benefit of sets?

pckhoi: It doesn't look like you're dealing with sets here. You only deal with sets if each row contains a finite number of unordered elements from an also finite set, which was the case for matching uids. First name and last name are definitely not sets in any sense. They are strings which are ordered lists of characters. You cannot use set operations on strings. String similarity functions such as Jaro Winkler's and Levenshtein's  are the ideal way to compare strings. Looks like you're already using one.

You have to look at how many pairs you're comparing and how many rows it is comparing per second. Both infos are available if you set `show_progress=True`. If the number of pairs is too high then you need a new index that is even more discrimative. If you don't know what other criteria that allow a pair to be matched but you do know other criteria that eliminate a pair then you can add a filter.

That and if you're only planning to match pairs perfectly (decision = 1) then there are little point in using datamatch. Just use pandas filtering instead which is guaranteed to be faster.

ayyubibrahimi: Thanks for the feedback!

data/raw/new_orleans_csc/appeal_hearing_pdfs/nopd_appeals_spacy_model.12_25_25.model/meta.json
```diff
@@ -0,0 +1,39 @@
...
+      "attorney_name",
+      "docket_number",
+      "incident_date",
+      "incident_summary",
+      "initial_action",
+      "nopd_incident_report_number",
+      "notification_of_action_date"
+    ]
+  }
+}
```
pckhoi, line 39: Rather keeping the model in Git, I think it's better to package the model using `python -m spacy package` and save it on Dropbox if you want to keep it. Or just git ignore it all if you're still improving the model.

extract/spacy_model.py
```diff
@@ -0,0 +1,101 @@
...
+
+nlp.begin_training()
+for itn in range(25):
+    random.shuffle(TRAINING_DATA)
+    losses = {}
+    for batch in spacy.util.minibatch(TRAINING_DATA, size=2):
+        texts = [text for text, entities in batch]
+        annotations = [entities for text, entities in batch]
+        nlp.update(texts, annotations, losses=losses, drop=0.3)
+    print(losses)
```
pckhoi, line 58: Try to put all configs in a `config.cfg` file, remove this script and invoke the training via `python -m spacy train config.cfg` instead. Convert training and evaluation data to spacy format using existing command line or if it's not available, write your own script and invoke them as necessary.

pckhoi:
If you still want to learn more about dependency and build order I recommend reading [GNU Make manual](https://www.gnu.org/software/make/manual/html_node/index.html#SEC_Contents) from beginning to end, which might seems daunting but I promise it is well worth the effort. After reading it you will know all the clever tricks employed in this repo.

extract/spacy_model.py
pckhoi: [Prodigy](https://prodi.gy/docs/recipes#data-to-spacy) seems to do the same thing as Doccano, but it can output data directly in Spacy format. Any reason why you did not use Prodigy?

data/raw/new_orleans_csc/appeal_hearing_pdfs/nopd_appeals_spacy_model.12_25_25.model/meta.json
pckhoi: In the new framework, models will be regarded as a kind of data, not a kind of code. Meaning it will not be kept in Git but in a data repository of some kind (Dropbox?) to be pulled like how you're pulling raw CSVs right now, or generated from scratch.

extract/spacy_model.py
ayyubibrahimi: Got it. Thanks.

ayyubibrahimi: Decided not to buy the Prodigy subscription because the team was already using Doccano in Tarak's workflow.

ayyubibrahimi:
Perfect. I'll start today. Thanks for the guidance!

data/raw/new_orleans_csc/appeal_hearing_pdfs/nopd_appeals_spacy_model.12_25_25.model/meta.json
ayyubibrahimi: All of my work on improving the model is local, so there is no need for this branch to exist for the moment. My intention with this branch was to learn from your initial feedback, as I found myself reaching a place in my progress where I'm finally deriving some utility from what I've produced so far. 

I'm going to close this pull request and remove the branch. We can continually revisit its progress as we continue our work on the new pipeline?

pckhoi: Sure

extract/spacy_model.py
pckhoi: I see

PR #223, opened 2022-01-02
enhancement/appeals

ayyubibrahimi:
- Generated an appeals table based on the three appeal_hearing_logs we've received from Baton Rouge PD, Louisiana State PD, and New Orleans PD.

### DIFF
--- a/data/datavalid.yml
+++ b/data/datavalid.yml
@@ -391,8 +391,8 @@ schemas:
       - name: uid
       - name: counsel
       - name: charging_supervisor
-      - name: resolution
-      - name: action
+      - name: appeal_disposition
+      - name: action_appealed
       - name: delay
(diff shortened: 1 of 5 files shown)


data/datavalid.yml
```diff
@@ -391,8 +391,8 @@ schemas:
       - name: uid
       - name: counsel
       - name: charging_supervisor
-      - name: resolution
-      - name: action
+      - name: appeal_disposition
+      - name: action_appealed
```
pckhoi, line 395: Please add schema for appeal hearing as well, and run datavalid afterward.

ayyubibrahimi: Thanks. Please review.

pckhoi (approved):
LGTM

PR #224, opened 2022-01-04
bug/wrgl-config.yaml

ayyubibrahimi:
removed match/pprr_brusly and match/pprr_st_tammany from wrgl/config. 

`wrgl diff --all `and `wrgl commit --all` runs smoothly when these two files, which are no longer being generated and used, are removed.


pckhoi (approved):
Lgtm

PR #225, opened 2022-01-05
enhancement/canonicalize_names function

ayyubibrahimi:
- New pull request opened with more appropriate PR name


ayyubibrahimi:
I've been having an impossible time creating a test for this function. Could you please provide a clue?

pckhoi:
@ayyubibrahimi it's fine, I will write the tests and push to this branch.

ayyubibrahimi:
Happy to write them myself, to learn. I can also invest my time in working on other tests.

pckhoi:
@ayyubibrahimi care for a live share session to work on the tests? I'm available this morning.

ayyubibrahimi:
That would be awesome. I can't tonight, unfortunately, I'm on my way out soon to see the new Matrix, but I can do any other day.

pckhoi:
Nice! Enjoy the movie! I'm available tomorrow.

ayyubibrahimi:
Let me know when is best for you and I'll send a link.

ayyubibrahimi:
Or do you want to share the link and I'll follow along?

ayyubibrahimi:
Update. 

Nearly there. UID successfully maps onto the given row but first_name and last_name is returning NaN for said row. Should be ready for review soon.

ayyubibrahimi:
@pckhoi - Please review when you have the time. Thanks!

lib/clean_test.py
```diff
@@ -36,3 +37,58 @@ def test_remove_future_dates(self):
                 columns=columns,
             ),
         )
+
+
+class FoatIntStr(unittest.TestCase):
```
pckhoi, line 42: All classes inherit from `unittest.TesetCase` should have "TestCase" suffix for consistency.

lib/clean_test.py
```diff
@@ -36,3 +37,58 @@ def test_remove_future_dates(self):
...
+        df = df.pipe(float_to_int_str, ["dates"])
+
+        dfa = pd.DataFrame(data)
+        dfa.loc[:, "dates"] = dfa.dates.astype(str)
+
+        df = df[["dates"]]
+
+        dfa = dfa[["dates"]]
+
+        assert_frame_equal(df, dfa)
```
pckhoi, line 57: - Rather than specifying `data["dates"]` as a list of strings and then converting them into floats, you can just specify them as a list of floats in the first place.
- This test could also test what happens when `float_to_int_str` processes more than one column.
- The primary reason why this function is even needed is because Pandas convert int columns with any `NA` value into float columns. You should insert some `np.nan` values and assert that they are turned into empty string.
- Since both `df` and `dfa` only have one column, `df = df[["dates"]]` and `dfa = dfa[["dates"]]` are no op.
- It is better if you specify the second argument to `assert_frame_equal` inline rather than elsewhere. That way people can see at a glance what is the expected output.

Here's how I would write this method just to solidify my previous points:
```python

class FloatToIntStrTestCase(unittest.TestCase):
    def test_float_to_int_str(self):
        """
        This is short but it tests multiple things:
        - float_to_int_str works on multiple columns
        - float_to_int_str only works on specified columns
        - float_to_int_str transforms float columns to integer strs
        - float_to_int_str leave int64 columns alone
        - float_to_int_str transforms "mixed" column (dtype object)
        """
        columns = ["id", "name", "age", "mixed"]
        assert_frame_equal(
            float_to_int_str(
                pd.DataFrame(
                    [
                        [3, "john", 27.0, 1973.0],
                        [4, "anne", 24.0, np.nan],
                        [5, "bill", np.nan, "abc"],
                    ],
                    columns=columns,
                ),
                ["id", "age", "mixed"],
            ),
            pd.DataFrame(
                [
                    [3, "john", "27", "1973"],
                    [4, "anne", "24", ""],
                    [5, "bill", "", "abc"],
                ],
                columns=columns,
            ),
        )
```

lib/clean.py
```diff
@@ -691,3 +691,33 @@ def strip_birth_date(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
...
+                )
+            ):
+                uid, first_name, last_name = (
+                    idx,
+                    row[first_name_column],
+                    row[last_name_column],
+                )
+        df.loc[df[uid_column].isin(cluster), uid_column] = uid
+        df.loc[df[uid_column] == uid, first_name_column] = first_name
+        df.loc[df[uid_column] == uid, last_name_column] = last_name
```
pckhoi, line 722: I think the last 2 rows should be:
```python
        df.loc[df[uid_column].isin(cluster), first_name_column] = first_name
        df.loc[df[uid_column].isin(cluster), last_name_column] = last_name
```
The original seems to be no op.

lib/clean_test.py
```diff
@@ -36,3 +37,58 @@ def test_remove_future_dates(self):
...
+
+        clusters = [
+            frozenset(
+                {"1e65c0807675ee3f25f6a6bf25eb121b", "db3d392b404754b2d4a127ca0922d2f8"}
+            )
+        ]
+
+        dfa = canonicalize_names(
+            pd.DataFrame(officers, columns=["uid", "first_name", "last_name"]), clusters
+        ).fillna("")
```
pckhoi, line 79: `clusters` should be a list of tuples instead of frozen sets. It will still work in this case because frozen sets are still iterable in the same way but it leaves people wondering why.

lib/clean_test.py
```diff
@@ -36,3 +37,58 @@ def test_remove_future_dates(self):
...
+            ["db3d392b404754b2d4a127ca0922d2f8", "", ""],
+            ["db3d392b404754b2d4a127ca0922d2f8", "patrick", "petermann"],
+            ["db3d392b404754b2d4a127ca0922d2f8", "patrick", "petermann"],
+        ]
+
+        dfb = pd.DataFrame(
+            canonicalized_officers, columns=["uid", "first_name", "last_name"]
+        )
+
+        assert_frame_equal(dfa, dfb)
```
pckhoi, line 94: After you have made the changes I suggested to `canonicalize_names`, you should test the following 2 cases:
- rows in the same cluster with shorter first name get their uid, first name, and last name replaced.
- rows in the same cluster with the same first name's length but shorter last name also get replaced.

Also not critical but I think `canonicalize_names` could be changed into `canonicalize_officers` since it also modifies uid.

match/new_orleans_so.py
```diff
@@ -39,28 +40,7 @@ def deduplicate_cprr_19_personnel(cprr):
...
-                    idx,
-                    row.first_name,
-                    row.last_name,
-                    row.middle_initial,
-                )
-        cprr.loc[cprr.uid.isin(cluster), "uid"] = uid
-        cprr.loc[cprr.uid == uid, "first_name"] = first_name
-        cprr.loc[cprr.uid == uid, "last_name"] = last_name
-        cprr.loc[cprr.uid == uid, "middle_initial"] = middle_initial
+    canonicalize_names(cprr, clusters)
```
pckhoi, line 43: Clearly we need to modify `canonicalize_names` to also accommodate `middle_initial` column, and possibly also `middle_name` column. The trick is to make sure the columns are all optional so for example if the frame only have `last_name` and `middle_initial`, it should still work without a hitch.

match/new_orleans_so.py
```diff
@@ -89,28 +69,7 @@ def deduplicate_cprr_20_personnel(cprr):
...
-                uid, first_name, last_name, middle_name = (
-                    idx,
-                    row.first_name,
-                    row.last_name,
-                    row.middle_name,
-                )
-        cprr.loc[cprr.uid.isin(cluster), "uid"] = uid
-        cprr.loc[cprr.uid == uid, "first_name"] = first_name
-        cprr.loc[cprr.uid == uid, "last_name"] = last_name
-        cprr.loc[cprr.uid == uid, "middle_name"] = middle_name
```
pckhoi, line 113: There really is a case with`middle_name` column involved. We could go even one step further, allowing the name columns be specified in a list e.g. `canonicalize_officers(cprr, id_column='uid', extra_columns=['first_name', 'last_name', 'middle_name'])`

match/new_orleans_so.py
ayyubibrahimi: Had this thought a while ago, but I forgot to bring it up. It might be worth standardizing `middle_initial `columns to `middle_name `columns as it's probable that we'll receive data that will allow us to replace initials with full middle names

lib/clean_test.py
ayyubibrahimi: I initially had a list of tuples, but changed it after reading the documentation for clusters in which it mentioned frozen sets.

lib/clean.py
ayyubibrahimi: Nice. I made that change when I was experimenting, but reverted back to what you originally produced.

lib/clean_test.py
ayyubibrahimi: Got it.

lib/clean_test.py
ayyubibrahimi: Cool.

match/new_orleans_so.py
ayyubibrahimi: The changes that I just pushed seem to allow for the function to work, regardless of a` middle_name `column, without a hitch.

match/new_orleans_so.py
pckhoi: I suppose we don't lose anything if we do that. @rajivsinclair ?

match/new_orleans_so.py
pckhoi: Not critical. We can modify it when we need to in the future.

lib/clean_test.py
```diff
@@ -36,3 +38,131 @@ def test_remove_future_dates(self):
...
+                        "first_name": "arthur",
+                        "middle_name": "hesse",
+                        "last_name": "schopenhaur",
+                    },
+                    {
+                        "uid": "233418697784e972144523ad8cc4ed9",
+                        "first_name": "",
+                        "middle_name": "",
+                        "last_name": "",
+                    },
```
pckhoi, line 165: It's strange that the name columns of non-canonical officers are empty. I don't think that's what we want, rather all columns including `uid` should be the same as the canonical one. Can you add some non-name columns too so we can assert that this function leaves other columns as is?

ayyubibrahimi: I also thought that was strange. I've added non-name columns to assert that this function leaves other columns as is. The code that you originally wrote (pasted below) also removed names and only replaced `uids`, which made me, believe that this was the intended output. Should this then be re-written to include the` name` columns in addition to the `uid` columns?

```
def canonicalize_names(
        df: pd.DataFrame, clusters: list[tuple], uid_column: str = 'uid',
        first_name_column: str = 'first_name', last_name_column: str = 'last_name'
) -> pd.DataFrame:
    for cluster in clusters:
        uid, first_name, last_name = None, '', ''
        for idx in cluster:
            row = df.loc[df[uid_column] == idx]
            if (
                uid is None
                or len(row[first_name_column]) > len(first_name)
                or (len(row[first_name_column]) == len(first_name) and len(row[last_name_column]) > len(last_name))
            ):
                uid, first_name, last_name = idx, row[first_name_column], row[last_name_column]
        df.loc[df[uid_column].isin(cluster), uid_column] = uid
        df.loc[df[uid_column] == uid, first_name_column] = first_name
        df.loc[df[uid_column] == uid, last_name_column] = last_name
    return df
```

pckhoi: Yes, it will be helpful in this instance to write the expected output first and then modify the code to match the output. Congrats! You have just discovered TDD. :)

Don't worry about my other suggestion, just focus on making the function correct for now.

ayyubibrahimi:
@pckhoi - the latest code is a working version of `canonicalize_officers` which passes both the test and works effectively when applied to `deduplicate_cprr_14_officers` and `deduplicate_cprr_14_officers` for Hammond PD, both of which do not contain a `middle_name` column. Per your last comment, I'm pushing this for your review before digging into the `middle_name`

lib/clean.py
```diff
@@ -691,3 +691,37 @@ def strip_birth_date(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
...
+                first_name = row[first_name_column]
+                first_names.append(first_name)
+
+                last_name = row[last_name_column]
+                last_names.append(last_name)
+
+                df.loc[df[uid_column].isin(cluster), uid_column] = uid
+                df.loc[df[uid_column].isin(cluster), first_name_column] = first_names
+                df.loc[df[uid_column].isin(cluster), last_name_column] = last_names
+    return df
```
pckhoi, line 727: So it looks like the only thing that changed is that you are assigning a list instead of a string to a column? That's really weird, I swear assigning a single string to an entire column used to work perfectly. Not that there's anything wrong with assigning a list of course. But other than that this code does have problems:
- Since the `first_names` and `last_names` are all reset in each new iteration, they are always going to be a 1-item list at the moment of column assignment.
- The column assignments:
```python
                df.loc[df[uid_column].isin(cluster), uid_column] = uid
                df.loc[df[uid_column].isin(cluster), first_name_column] = first_names
                df.loc[df[uid_column].isin(cluster), last_name_column] = last_names
```
should be done outside of the second for loop. It is very strange that this isn't crashing for any cluster that has more than 1 uid. There is a dimensions miss-match: for every n-item cluster, you're assigning:
- a uid string to an (n x 1) uid series, which should work of course.
- a list of first names which only ever has 1 item (1 x 1) to an (n x 1) first name series, which strangely doesn't crash.  I guess a list of 1 string is treated as a single string?
- (1 x 1) last names to an (n x 1) last name series, which works similarly to the first names.

All in all, I think you should get rid of `first_names` and `last_names` as assigning a single string to a series is possible as shown with the uid assignment. I suspect the previous implementation has a bug that somehow made the `first_name` and `last_name` empty. And please move the column assignments out of the second for loop.

ayyubibrahimi: The code I pushed seems to have fixed the bug that made the `first_name` and `last_name` empty.

pckhoi: What is the purpose of `' '.join(row[last_name_column])`? I thought the name cells are all single strings, not iterables.

ayyubibrahimi: Why it works? No idea. I was hoping that you could tell me. 

Running 

```
uid = idx
first_name = row[first_name_column]
last_name = row[last_name_column]
```
returns the correct uid, with but does not return the `first_name` and `last_name` for the row(s) whose `uid `has been modified.

Running 

```
uid = idx
first_name = ' '.join(row[first_name_column])
last_name = ' '.join(row[last_name_column])
```
returns the `uid`,  in addition to the `first_name` and ` last_name` for all rows, including the row(s) which has been modified.

ayyubibrahimi: In other news, I've decided that this test is insufficient after reviewing the cluster sizes in `match/new_orleans_so` i.e., there are 6 individuals who are all likely to be the same person whereas this test has been limited to two individuals.

pckhoi: Well, one should always keep all possible scenarios in mind when writing tests and make sure the tests sufficiently address all edge cases. Do you mind if I take a stab at writing this function and its tests?

pckhoi:
@ayyubibrahimi I'm merging code and improving tests on this branch. I hope you don't mind me touching this branch but I feel I should for having neglected it for so long.

ayyubibrahimi:
Misunderstanding. I thought that based on your last message you were going to take over. Please go for it.

pckhoi (approved):
LGTM

PR #226, opened 2022-01-06
Enhancement/allegations_rayne_pd_2019_2020

ayyubibrahimi:


.wrgl/config.yaml
```diff
@@ -459,6 +463,10 @@ remote:
     url: https://hub.wrgl.co/api/users/ipno/repos/event-acadia-so
     fetch:
       - +refs/heads/*:refs/remotes/event-acadia-so/*
+  event-rayne-pd:
+    url: https://hub.wrgl.co/api/users/ipno/repos/event-rayne-pd
+    fetch:
+      - +refs/heads/*:refs/remotes/event-rayne-pd/*
```
pckhoi, line 469: There should be an entry for `event-rayne-pd` in the `branch` section no?

ayyubibrahimi: Just checked and its there. There was a duplicate event entry in the remote section which I just removed.

pckhoi (approved):
LGTM

PR #229, opened 2022-01-11
enhancement/post

ayyubibrahimi:
@pckhoi - Rajiv recommended fusing `fuse/personnel_post.csv` into `fuse/personnel.csv`, rather than having post personnel exist as separate table and repo. Please review

### DIFF
--- a/fuse/all.py
+++ b/fuse/all.py
@@ -64,6 +64,7 @@ def fuse_personnel():
                 pd.read_csv(data_file_path("fuse/per_terrebonne_so.csv")),
                 pd.read_csv(data_file_path("fuse/per_jefferson_so.csv")),
                 pd.read_csv(data_file_path("fuse/per_acadia_so.csv")),
+                pd.read_csv(data_file_path("fuse/per_post.csv")),
             ]
         )
     ).sort_values("uid", ignore_index=True)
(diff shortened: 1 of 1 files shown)


ayyubibrahimi:
We discussed the issue of conflicting `hire_dates`. We could concat post events only when there is no `hire_date` already in the common events table?

pckhoi:
Aren't we only including agencies that don't have any other data sources in post events? Which agencies are included in the post events table?

The hire date conflicts issue is something that is unlikely to resolve anytime soon. The LLEAD website will never display conflicting data so is it practical to keep multiple conflicting hire dates?

ayyubibrahimi:
Sorry. I was only thinking about the concatenation prior to the creation of the `post_event_df`

Yes. the `post_events_df `only has data sources from agencies for which we have no other data. 

```
post_event_df = post_event_df[~post_event_df["agency"].isin(event_df["agency"])]
post_event_df.to_csv(data_file_path("fuse/event_post.csv"), index=False)
```
Your idea was to concatenate this `post_event_df` with the `event_df `?

pckhoi:
Then there is no problem if we simply concatenate post events with the rest. Since there are no other data sources, there could be no conflicts. Please go ahead.

ayyubibrahimi:
Closing this PR in favor of PR #230

PR #233, opened 2022-01-13
enhancement/pprr_baker_pd

ayyubibrahimi:


ayyubibrahimi:
Forgot to run names through the `clean_names` function as we discussed should be standard practice from now on. This is now ready for review.

clean/baker_pd_pprr.py
```diff
@@ -0,0 +1,67 @@
...
+        pd.read_csv(data_file_path("raw/baker_pd/baker_csd_pprr_2010_2020.csv"))
+        .pipe(clean_column_names)
+        .drop(columns=["annual_salary", "monthly_salary", "hourly_salary"])
+        .rename(columns={"date_of_hire": "hire_date"})
+        .pipe(clean_rank_desc)
+        .pipe(clean_resignation_date)
+        .pipe(clean_dates, ["hire_date", "resignation_date"])
+        .pipe(standardize_desc_cols, ["race", "sex"])
+        .pipe(clean_races, ["race"])
+        .pipe(clean_sexes, ["sex"])
```
pckhoi, line 56: There's no harm in running `race` and `sex` through `standardize_desc_cols` but I think it is a little excessive. `clean_races` and `clean_sexes` should be robust enough that there's no need to put `standardize_desc_cols` before it.

pckhoi (approved):
LGTM

PR #243, opened 2022-01-25
bugs/miscellaneous

ayyubibrahimi:


data/datavalid.yml
```diff
@@ -402,6 +402,15 @@ schemas:
...
       - name: motions
+      - name: appeal_disposition_year
+      - name: appeal_disposition_month
+      - name: appeal_disposition_day 
+      - name: appeal_hearing_year
+      - name: appeal_hearing_month
+      - name: appeal_hearing_day
+      - name: appeal_hearing_2_year 
+      - name: appeal_hearing_2_month
+      - name: appeal_hearing_2_day
```
pckhoi, line 413: Can we replace these columns with "appeal_disposition", "appeal_hearing_1", "appeal_hearing_2" events instead?

ayyubibrahimi: Replaced.

data/datavalid.yml
```diff
@@ -402,15 +398,9 @@ schemas:
...
-      - name: appeal_disposition_day 
-      - name: appeal_hearing_year
-      - name: appeal_hearing_month
-      - name: appeal_hearing_day
-      - name: appeal_hearing_2_year 
-      - name: appeal_hearing_2_month
-      - name: appeal_hearing_2_day
+      - name: appeal_disposition_date
+      - name: appeal_hearing_date
+      - name: appeal_hearing_2_date
```
pckhoi, line 403: Since we already have the events, we don't need these date columns anymore do we?

ayyubibrahimi: Good point. @rajivsinclair I made these changes based on our discussion yesterday re the exclusion of dates in the appeals table. Thoughts?

rajivsinclair:
@ayyubibrahimi I think we should only store the date of any event in the event table.

data/datavalid.yml
rajivsinclair: Sorry for any confusion. Are there more than one type of date for each appeal event? If so, only the primary date field should appear in the event table. The other date fields could be stored here in this table and attributes of the event.

rajivsinclair: Let's discuss this further.

rajivsinclair: Sorry for the confusion/back-and-forth on this, but upon further discussion and thought, we decided that each date should be a separate event (with a different `event_kind`) and these events should be linked together by a common `appeal_uid`. This way we will be consistent with the existing method for handling of multiple dates that are already associated with a single record as with `allegation` data.

ayyubibrahimi:
You'll notice that I had a deep back and forth between how to treat rows missing `first_name` and `last_name`. In the end, I decided to drop them completely as you and Rajiv initially suggested.

pckhoi:
I see that you are dropping rows without `first_name` and `last_name`. I thought I advocated for simply setting the `uid` column to empty string instead of dropping them? Is there a misunderstanding?

ayyubibrahimi:
Only a slight one. We began by dropping them, which you and Rajiv agreed with. After I had a difference of opinion, you advocated for simply setting the `uid` to an empty string instead of dropping them. Yesterday, I came to the conclusion that your initial opinion, that we drop them, made the most sense.

clean/baton_rouge_pd_cprr.py
```diff
@@ -504,15 +504,12 @@ def split_name_21(df):
...
-    )
-    df.loc[:, "middle_initial"] = (
-        names.loc[:, 2].str.strip().fillna("").map(lambda s: "" if len(s) > 2 else s)
+    df.loc[:, "first_name"] = names[1].str.strip().fillna("")
+    df.loc[:, "last_name"] = (
+        names[3].str.strip().fillna("").str.replace(r"\,", " ", regex=True)
     )
-    return df.drop(columns=["name"])
+    df.loc[:, "middle_name"] = names[2]
+    return df[~((df.first_name == "") & (df.last_name == ""))].drop(columns="name")
```
pckhoi, line 512: I don't think the `.drop(columns="name")` bit does anything.

pckhoi (approved):
LGTM

clean/baton_rouge_pd_cprr.py
ayyubibrahimi: It seems to be correctly dropping the `name` column after splitting said column into `first_name`, `middle_name`, and `last_name`. I'm going to go ahead and merge.

pckhoi: @ayyubibrahimi I tried this at local and doesn't look like you can drop column on a subset of rows:
```python
>>> import pandas as pd
>>> df = pd.DataFrame([[1,2],[3,4]], columns=['a','b'])
>>> df
   a  b
0  1  2
1  3  4
>>> df.loc[0]
a    1
b    2
Name: 0, dtype: int64
>>> df.loc[0].drop(columns='a')
a    1
b    2
Name: 0, dtype: int64
>>> df
   a  b
0  1  2
1  3  4
```
But maybe I'm missing something.

ayyubibrahimi: This is my output: 

1) `drop(columns="name") `is included in the return statement of the `split_name_21 function`
2) after dropping `drop(columns="name")` from the `split_name_21` function in the baton_rouge_pd_cprr.py script

```
>>> from importlib import reload
>>> from clean import baton_rouge_pd_cprr as n
>>> df = n.clean_21()
standardize_from_lookup_table: unmatched sequences:
  set()
>>> df.columns
Index(['received', 'action', 'disposition', 'complainant',
       'investigation_status', 'tracking_number', 'receive_year',      
       'receive_month', 'receive_day', 'occur_year', 'occur_month',    
       'occur_day', 'allegation', 'department_code', 'department_desc',
       'rank_desc', 'first_name', 'last_name', 'middle_name', 'division_desc',
       'agency', 'data_production_year', 'uid', 'allegation_uid'],
      dtype='object')
>>> reload(n)
<module 'clean.baton_rouge_pd_cprr' from '/mnt/c/Users/ayyubi/lleads/processing/clean/baton_rouge_pd_cprr.py'>
>>> df = n.clean_21()
standardize_from_lookup_table: unmatched sequences:
  set()
>>> df.columns
Index(['received', 'action', 'disposition', 'complainant',
       'investigation_status', 'tracking_number', 'receive_year',
       'receive_month', 'receive_day', 'occur_year', 'occur_month',
       'occur_day', 'allegation', 'name', 'department_code', 'department_desc',
       'rank_desc', 'first_name', 'last_name', 'middle_name', 'division_desc',
       'agency', 'data_production_year', 'uid', 'allegation_uid'],
      dtype='object')
```

Thoughts?

pckhoi: It seems that the reason is you were not using `.loc`. It's fine then. Sorry for the false alarm :)

ayyubibrahimi: Should I be using `.loc` here? Happy to make the change.

pckhoi: Oh no, I would say the only time when `.loc` must be used is when assigning. Not using `.loc` in this case does make it shorter. It is up to your judgment then.

rajivsinclair:
@ayyubibrahimi does this implementation include storing the appeal's the secondary date as a separate event (with a distinct event `kind`)?

ayyubibrahimi:
Yes.

rajivsinclair:
Thanks @ayyubibrahimi what is the `kind` for these second dates? I couldn't find any when I looked. 

Also, what do you think we should display as a label for this field on the appeal hearing card?

ayyubibrahimi:
`appeal_hearing_2` is the name of the event. 

What is the label for the first `appeal_hearing` event?

rajivsinclair:
There isn't actually a label for the main appeal_hearing date inside the card, it appears just above the card and it is used to determine where the card will appear on the timeline. I think we've agreed that this secondary date (which is a separate event) is just going to appear as another row inside of that card (which will need a label), see attached screenshot: 
image

PR #246, opened 2022-01-26
Enhancement/brady_lists (baton rouge da; new orleans da; ouachita da; tangipahoa da)

ayyubibrahimi:


clean/tangipahoa_da.py
```diff
@@ -3,19 +3,25 @@
...
 
 def assign_agency(df):
-    df.loc[:, "agency"] = ""
+    df.loc[
+        (df.last_name == "gemar") & (df.first_name == "rodney"), "agency"
+    ] = "Hammond PD"
+
+    df.loc[
+        (df.last_name == "hampton") & (df.first_name == "mark"), "agency"
+    ] = "Hammond PD"
```
pckhoi, line 24: What is "Hammond PD" doing here?

fuse/baton_rouge_da.py
```diff
@@ -0,0 +1,44 @@
...
+        ["uid"],
+    )
+    return builder.to_frame()
+
+
+if __name__ == "__main__":
+    brady = pd.read_csv(data_file_path("clean/cprr_baton_rouge_da_2021.csv"))
+    agency = brady.agency[0]
+    post = load_for_agency("clean/pprr_post_2020_11_06.csv", agency)
+    per = fuse_personnel(brady, post)
```
pckhoi, line 38: I saw 3 personnel datasets for Baton Rouge DA. Why do you take roster data from Brady and POST instead?

fuse/new_orleans_da.py
```diff
@@ -0,0 +1,57 @@
...
+    brady = pd.read_csv(data_file_path("match/cprr_new_orleans_da_2021.csv"))
+    agency = brady.agency[0]
+    post = load_for_agency("clean/pprr_post_2020_11_06.csv", agency)
+    per = fuse_personnel(brady, post)
+    complaints = rearrange_allegation_columns(brady)
+    event = fuse_events(brady, post)
+    complaints.to_csv(data_file_path("fuse/com_new_orleans_da.csv"), index=False)
+    per.to_csv(data_file_path("fuse/per_new_orleans_da.csv"), index=False)
+    event.to_csv(data_file_path("fuse/event_new_orleans_da.csv"), index=False)
+    brady.to_csv(data_file_path("fuse/brady_new_orleans_da.csv"), index=False)
```
pckhoi, line 57: Looks like the Brady list is output to both allegation and brady list tables. Seems a little redundant?

fuse/baton_rouge_da.py
ayyubibrahimi: The officers listed on a brady list aren't employed by the DA so they wont' show up on the personnel dataset. The brady list is a list created by the DA office of officers, who are employed either by a PD or SO.

fuse/new_orleans_da.py
ayyubibrahimi: I think so as well. I wasn't sure at first, but yes, the brady list tables are really just allegation tables, but of a special type that are worth demarcation, therefore, it is redundant. I'll make that change.

clean/tangipahoa_da.py
ayyubibrahimi: DA offices are entities completely separate from PD or SO's, therefore, multiple law enforcement agencies can end up in one brady list. 

For example, the Baton Rouge DA brady list includes officers from Baton Rouge PD, Baton Rouge SO, Louisiana State PD, LSU University PD and the Department of Public Saffety. 

The brady list for Tangipahoa DA includes officers from Hammond PD and Tangipahoa SO because Hammond PD is an agency located in the city of Hammond which is itself located in the Tangipahoa region of Louisiana.

fuse/new_orleans_da.py
ayyubibrahimi: These tables probably shouldn't be fused into the personnel table either, because they aren't employed by the DA. Does that sound right to you?

ayyubibrahimi:
I removed the brady lists from the allegations table. 

Additionally, because officers can be from various agencies, and because said officers even might work for a different agency today, I thought that it might make more sense to match these brady lists only with POST. What do you think?

fuse/baton_rouge_da.py
pckhoi: Aren't we supposed to save officers who are employed by the DA in the file `fuse/per_baton_rouge_da.csv`?

fuse/new_orleans_da.py
pckhoi: When you said individuals who are associated with the DA's office, are you referring to individuals on the brady list or elsewhere? Are there a third type of individuals who are neither employed by the DA nor appear on the brady list?

But yeah, my impression is that only employed individuals should appear on the personnel table in a way that the `agency` column in the events associated with that individual should be the employer.

So the rows on brady lists should really be associated with the agencies that employed them rather than the DA I think. That way it will not break the semantic of the `agency` column and you can still show both the place that employs an officer and on which brady list, from which DA said officer appears in.

@rajivsinclair @lsukaristone what do you think?

ayyubibrahimi: That sounds right to me.

clean/tangipahoa_da.py
```diff
@@ -24,12 +30,15 @@ def clean():
         pd.read_csv(data_file_path("raw/tangipahoa_da/tangipahoa_da_brady_2021.csv"))
         .pipe(clean_column_names)
         .pipe(split_name)
+        .pipe(set_values, {"agency": "Tangipahoa SO", "source_agency": "Tangipahoa DA"})
         .pipe(assign_agency)
         .pipe(gen_uid, ["first_name", "last_name", "agency"])
+        .pipe(gen_uid, ["uid", "source_agency"], "allegation_uid")
+        .pipe(gen_uid, ["uid", "source_agency"], "brady_uid")
```
pckhoi, line 37: Here, the `allegation_uid` is exactly the same as `source_agency`. Is there no other field we can rely on? If not perhaps we can generate the `allegation_uid` from `uid` and `agency` instead.

data/datavalid.yml
```diff
@@ -456,5 +457,18 @@ schemas:
...
+        description: >
+          unique brady uid 
+      - name: uid
+      - name: allegation
+      - name: disposition
+      - name: tracking_number
+      - name: action
+      - name: agency
+      - name: source_agency
+      - name: charging_agency
```
pckhoi, line 472: To normalize even more, I suggest putting `allegation`, `disposition`, `tracking_number`, `action`, and perhaps `charging_agency` as well in allegation table. And then link to brady list record via `allegation_uid`.

pckhoi: If you decide to follow that route then the best place to extract/merge rows is probably the fuse step.

match/new_orleans_pd.py
```diff
@@ -205,6 +205,42 @@ def match_stop_and_search_to_pprr(sas, pprr_ipm):
...
+        data_file_path(
+            "match/nopd_brady_2021_v_pprr_new_orleans_pd_1946_2018.xlsx"
+        ),
+        decision,
+    )
+    matches = matcher.get_index_pairs_within_thresholds(lower_bound=decision)
+    match_dict = dict(matches)
+
+    sas.loc[:, "uid"] = sas.uid.map(lambda x: match_dict.get(x, x))
+    return sas
```
pckhoi, line 241: I think `sas` is not defined in this function.

pckhoi: If you decide to match brady list's `uid` to post's `uid` then this script is not doing that.

match/new_orleans_pd.py
```diff
@@ -215,6 +251,8 @@ def match_stop_and_search_to_pprr(sas, pprr_ipm):
     award = pd.read_csv(data_file_path("clean/award_new_orleans_pd_2016_2021.csv"))
     lprr = pd.read_csv(data_file_path("clean/lprr_new_orleans_csc_2000_2016.csv"))
     sas = pd.read_csv(data_file_path("clean/sas_new_orleans_pd_2017_2021.csv"))
+    brady = pd.read_csv(data_file_path('clean/cprr_new_orleans_da_2021.csv'))
+    brady_df = match_brady_to_cprr(brady, pprr_ipm)
```
pckhoi, line 255: The name should be `match_brady_to_pprr`.

match/ouachita_da.py
```diff
@@ -7,41 +7,41 @@
...
+        ColumnsIndex(["fc", "lc"]),
         {
             "last_name": JaroWinklerSimilarity(),
             "first_name": JaroWinklerSimilarity(),
         },
         dfa,
         dfb,
     )
-    decision = 1
+    decision = 0
```
pckhoi, line 30: This is a very low decision. Keep in mind that score below 0.5 is considered more likely to be different records.

pckhoi (commented):
Sorry for this very late review. Somehow this PR escaped my attention.

Regarding whether you should match with Post data only. I think it comes down to purpose. If you only plan to use this together with records from POST then it is fine. However, if you plan to refer to `allegation` and `personnel` as well then I think it's better to normalize and match the `allegation_uid` and `uid` to those two tables respectively.

ayyubibrahimi:
I've removed the `allegation_uid` completely because these allegation data are different from the allegation data that we have received in the past, in that we are receiving it from the DA's office, a secondary source. Although some of the Brady lists contain allegations data, a Brady list is ultimately just a list of officer names that the DA produces for internal use. I expect that most Brady lists that we receive in the future will only contain officer names. For this reason, I've my edits in the context of treating these lists solely as a list of names.

pckhoi:
Alright, are you still going to match with POST's `uid` or are you going to match with personnel's `uid`?

ayyubibrahimi:
I came to the conclusion that it makes the most sense to match solely with personnel's `uid`. Thoughts?

pckhoi:
I think that choice makes the most sense as it integrate with everything else.

ayyubibrahimi:
Awesome. Everything else should be in order. Please review at your convenience.

.wrgl/config.yaml
```diff
@@ -1506,6 +1522,16 @@ branch:
...
+  event-new-orleans-da:
+    remote: event-new-orleans-da
+    merge: refs/heads/main
+    file: data/fuse/event_new_orleans_da.csv
+    primaryKey:
+      - event_uid
+  event-ouachita-da:
+    remote: event-ouachita-da
+    merge: refs/heads/main
+    file: data/fuse/event_ouachita_da.csv
```
pckhoi, line 1534: Missing primary key.

data/raw.dvc
```diff
@@ -1,5 +1,5 @@
 outs:
-- md5: dd63c99911d973223fa2a74e2f6e077c.dir
-  size: 303527593
-  nfiles: 154
+- md5: 2d085efbf8595e42cf6c1ab60dd12f47.dir
+  size: 307347037
+  nfiles: 155
```
pckhoi, line 4: I saw in another PR where nfiles is 156. Please be careful when merging, or raw data files could disappear.

.wrgl/config.yaml
ayyubibrahimi: Thank you. Sorry. Won't happen again.

data/raw.dvc
ayyubibrahimi: Got it.

pckhoi (approved):
LGTM

PR #258, opened 2022-02-28
Enhancement/nopd

ayyubibrahimi:
Hey @pckhoi - Here is a use of force table that we received from NOPD which is of higher quality than the table that we received from IPM. As is, the clean script runs pretty slowly, however, I've yet to figure out a more efficient way to split out the columns and rows. Thoughts?


clean/new_orleans_pd_uof.py
```diff
@@ -0,0 +1,732 @@
...
+        .str.replace(r"(\w+)\|", r"\1 |", regex=True)
+        .str.replace(r"(\w+);", r"\1 ;", regex=True)
+        .str.replace(
+            r"(\w+) \| (\w+) ; (\w+) \| (\w+) ; (\w+) \| (\w+) ; (\w+) \| (\w+) ; "
+            r"(\w+) \| (\w+) ; (\w+) \| (\w+) ; (\w+) \| (\w+) ; (\w+) \| (\w+) ; "
+            r"(\w+) \| (\w+) ; (\w+) \| (\w+) ; (\w+) ?\|? ?(\w+)?",
+            r"(\1 | \3 | \5 | \7 | \9 | \11 | \13 | \15 | \17 | \19 | \21) & "
+            r"(\2 | \4 | \6 | \8 | \10 | \12 | \14 | \16 | \18 | \20 | \22)",
+            regex=True,
+        )
```
pckhoi, line 62: This is very hard to understand. And regular expression is actually very slow compared to other string manipulation methods. I think you should rewrite `join_citizen_columns` to not rely on regular expression so much anymore.

pckhoi (commented):
So I opened https://github.com/ppact/processing/pull/259 to make `deba` work in a notebook which also introduce the `timed` accessor. Using the `timed` accessor I identified `join_officer_columns` as the most time consuming:
```
pipe(clean_column_names) (0.00s)
pipe(join_citizen_columns) (0.28s)
pipe(join_officer_columns) (730.71s)
pipe(split_rows_with_multiple_citizens) (0.02s)
pipe(split_rows_with_multiple_officers) (0.02s)
pipe(split_officer_column) (0.26s)
pipe(split_citizen_column) (0.31s)
pipe(clean_tracking_number) (0.02s)
pipe(clean_dates) (0.05s)
pipe(clean_originating_bureau) (0.01s)
pipe(clean_division_level) (0.03s)
pipe(clean_division) (0.04s)
pipe(clean_unit) (0.05s)
pipe(clean_working_status) (0.01s)
pipe(clean_shift) (0.01s)
pipe(clean_disposition) (0.01s)
pipe(clean_service_type) (0.02s)
pipe(clean_light_condition) (0.01s)
pipe(clean_weather_condition) (0.01s)
pipe(clean_use_of_force_reason) (0.03s)
pipe(standardize_desc_cols) (0.01s)
pipe(set_values) (0.00s)
pipe(gen_uid) (0.44s)
pipe(gen_uid) (0.32s)
```
I recommend rewriting that function from scratch because it is too opaque and rely too much on regular expression. Regular expression is very very slow, so use it sparingly.

data/datavalid.yml
```diff
@@ -363,24 +346,17 @@ schemas:
           - hispanic
           - native american
           - asian / pacific islander
+          - asian
```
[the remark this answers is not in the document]
ayyubibrahimi: Ok.

raw_datasets.json
```diff
@@ -280,5 +280,8 @@
   ],
   "eunice_pd": [
     "https://www.dropbox.com/s/vogk226khgov0vk/eunice_pd_cprr_2019_2021.csv?dl=1"
+  ],
+  "new_orleans_pd": [
+    "https://www.dropbox.com/s/7pzejv0xojp0os9/new_orleans_pd_uof_2016_2021.csv?dl=1"
```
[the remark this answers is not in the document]
ayyubibrahimi: Ok.

clean/new_orleans_pd_uof.py
ayyubibrahimi: Ok.

pckhoi:
@ayyubibrahimi here's an idea. Instead of joining multiple columns into one string column, how about joining the columns into one tuple column? Tuple manipulation is light year faster than regular expression. Also happy to rewrite this part if you want me to.

ayyubibrahimi:
I'll give it a shot. Thanks for the suggestion!

ayyubibrahimi:
Having some difficulty. Please go for it and I'll learn by review.

pckhoi:
@ayyubibrahimi in this script you seem to want to create one row per officer x subject combination. E.g. if a tracking number has 2 officers and 3 subjects then you want to create 6 rows for this tracking number in the final result. However I don't think this is the right approach.

There is a relationship between each subject and the report, likewise there is a relationship between each mentioned officer and the report. However, there is no explicit relationship between each officer and subject pair. For example, officer A and B has `Use of force type` "Take down" and "Taser" on citizen C and D. It is unknown whether officer A "Take down" both C and D, or just C, or just D. Therefore I think we're making too much assumption when joining each officer with each subject.

If you want to preserve the data integrity then just leave those columns alone and extract the numbers that are actually gonna be used on the LLEAD website such as "number of subjects involved" or "number of officers involved", or any kind of information from officers and subjects really. I just think it's not correct to join officer columns and subject columns. For example, if you start having records implying officer A taser subject C where in the original record, there was multiple officers and subjects involved, then I think we're misrepresenting the data.

@rajivsinclair I think it's a good time to start discussing what should be done with UOF records that have multiple officers and/or multiple subjects. Specifically what data should we display on the website.

pckhoi:
You can say that without making any assumptions nor joining the tables. So, staying true to the spirit of the `clean` step, I recommend splitting the `use_of_force` table into 3 tables: `use_of_force`, `uof_officers`, and `uof_subjects`. You can modify the function that I gave to produce a new dataframe instead. Just please don't join the officers and the subjects.

ayyubibrahimi:
Got it. Thanks.

ayyubibrahimi:
Just pushed the edits. Is this what you had in mind? @pckhoi 

I was curious as to if I should leave the `subject/citizen` columns as is in the `use_of_force` table, i..e, unsplit, because 1) we aren't generating a `citizen_uid` and 2) citizen data has been extracted from the `uof_kenner_pd_2005_2021.csv` table into the `use_of_force` table

clean/new_orleans_pd_uof.py
```diff
@@ -0,0 +1,552 @@
...
+        .dropna(subset=["uof_uid"])
+        .drop_duplicates(subset=["uof_uid"])
+    )
+    return df
+
+
+if __name__ == "__main__":
+    uof_citizen = clean_citizen()
+    uof_officer = clean_officer()
+    uof = clean_uof()
```
pckhoi, line 549: This is not super efficient. Ideally every column should only be dropped, or renamed, or cleaned only once. So something like this is more ideal:
```python
uof = clean_uof()
# extract citizen columns into a separate dataframe and remove citizen columns from uof
uof_citizen, uof = extract_citizen(uof)
uof_officer, uof = extract_officer(uof)

Comment (ayyubibrahimi) [reply clean/new_orleans_pd_uof.py:549]: Got it. Thanks.

File: clean/new_orleans_pd_uof.py [hunk 5]
```diff
@@ -0,0 +1,552 @@
...
+            "citizen_hospitalized",
+            "citizen_injured",
+            "citizen_influencing_factors",
+            "citizen_distance_from_officer",
+            "citizen_age",
+            "citizen_build",
+            "citizen_height",
+            "citizen_arrested",
+            "citizen_arrest_charges",
+        ]
```
pckhoi, line 52: Just plug the `citizen_columns` here instead of repeating them all again.

clean/new_orleans_pd_uof.py
```diff
@@ -0,0 +1,552 @@
...
+    )
+
+    df.loc[:, "citizen"] = df.citizen.apply(pd.DataFrame)
+    df = df.drop("citizen", axis=1).join(
+        pd.concat(df["citizen"].values, keys=df.index).droplevel(1)
+    )
```
pckhoi, line 58: This is not very efficient for the following reasons:
- You're still creating dictionaries but if we're going for a DataFrame, creating lists is more efficient. We already have the column names in a list after all.
- You're constructing a data frame for each cell, and then concat them. `pd.DataFrame` do a lot of preprocessing such as indexing data upon creation so it is more efficient to concatenate all records into a big record list and then call `pd.DataFrame` only once.
- No need to join with all the existing columns. We only need to tack on the `tracking_number`.
```python
# return a list of records for each row, notice that we also prepend
# tracking_number to each record.
def create_citizen_lists(row: pd.Series):
    d = row.loc[citizen_columns].loc[row.notna()].to_dict()
    return [
        [row.tracking_number] + [d[k][i] for k in d.keys()]
        for i in range(max(len(v) for v in d.values()))
    ]

citizen_series = df.apply(create_citizen_lists, axis=1)
citizen_df = pd.DataFrame(
    # double list comprehension is the fastest way to concat a list of lists
    [element for list_ in citizen_series for element in list_],
    columns=['tracking_number']+citizen_columns,
)
```

fuse/new_orleans_pd.py
```diff
@@ -162,20 +150,26 @@ def fuse_events(pprr_ipm, pprr_csd, cprr, uof, award, lprr, sas):
...
     cprr = pd.read_csv(deba.data("clean/cprr_new_orleans_pd_1931_2020.csv"))
     actions = pd.read_csv(deba.data("clean/cprr_actions_new_orleans_pd_1931_2020.csv"))
-    uof = pd.read_csv(deba.data("clean/uof_new_orleans_pd_2012_2019.csv"))
+    uof_officers = pd.read_csv(deba.data("clean/uof_officers_new_orleans_pd_2016_2021.csv"))
+    uof_citizens = pd.read_csv(deba.data("clean/uof_citizens_new_orleans_pd_2016_2021.csv"))
+    uof = pd.read_csv(deba.data("clean/uof_new_orleans_pd_2016_2021.csv"))
+    uof_dfs = [uof_officers, uof_citizens, uof]
+    uof = reduce(
+        lambda left, right: pd.merge(left, right, on="tracking_number"), uof_dfs
+    ).drop_duplicates(subset=["uof_uid"], keep="first")
```
pckhoi, line 159: Please don't join or merge these tables. Here's what I think should be in the table:
- `uof_officers`: has `tracking_number` and all the officer columns.
- `uof_citizens`: has `tracking_number` and all the citizens columns.
- `uof`: has `tracking_number` of course and all the remaining columns.

No column should be repeated across tables.

pckhoi:
@ayyubibrahimi since we're start having officers and subjects in the same dataset, we cannot go back to how it was anymore. Like I have iterated many times already, we should avoid merging officers and subjects. Going forward, all uof citizens data should be read off of `uof_citizen` which is an actual final output table, not just intermediary. We can generate and put the `citizen_uid` in that table. And we should go back and change Kenner PD scripts to adhere to the new structure too.

ayyubibrahimi:
Apologies for misunderstanding. I'll make the changes. Thanks.

pckhoi:
Apologies for not being clearer :)

ayyubibrahimi:
@pckhoi - Is this more what you had in mind?

clean/new_orleans_pd_uof.py
```diff
@@ -32,14 +32,8 @@ def create_uof_lists(row: pd.Series):
...
-        columns=[
-            "use_of_force_type",
-            "use_of_force_level",
-            "use_of_force_effective",
-            "officer_injured",
-        ]
-    )
+
+    df = df.drop(columns=uof_columns)
     df = pd.merge(df, uof_df, on="tracking_number", how="outer")
```
ayyubibrahimi, line 37: @pckhoi - I'm not sure as to if this is the correct way to re-join the `uof_df` with the original `df` after splitting the rows with multiple uof values.

pckhoi: Shouldn't the `use_of_force_*` columns be in the `uof_officers` table? They are specific to each officer are they not?

fuse/new_orleans_pd.py
```diff
@@ -162,21 +150,25 @@ def fuse_events(pprr_ipm, pprr_csd, cprr, uof, award, lprr, sas):
...
+    uof_df = rearrange_use_of_force(uof)
+    citizen_df = rearrange_citizen_columns(uof_citizens)
     complaints.to_csv(deba.data("fuse/com_new_orleans_pd.csv"), index=False)
-    use_of_force.to_csv(deba.data("fuse/uof_new_orleans_pd.csv"), index=False)
+    uof_df.to_csv(deba.data("fuse/uof_new_orleans_pd.csv"), index=False)
     personnel.to_csv(deba.data("fuse/per_new_orleans_pd.csv"), index=False)
     events_df.to_csv(deba.data("fuse/event_new_orleans_pd.csv"), index=False)
     lprr_df.to_csv(deba.data("fuse/app_new_orleans_csc.csv"), index=False)
     sas_df.to_csv(deba.data("fuse/sas_new_orleans_pd.csv"), index=False)
+    citizen_df.to_csv(deba.data("fuse/citizen_new_orleans_pd.csv"))
```
pckhoi, line 174: I think the file name should have "uof_citizen_" and the prefix.

fuse/new_orleans_pd.py
```diff
@@ -162,21 +150,25 @@ def fuse_events(pprr_ipm, pprr_csd, cprr, uof, award, lprr, sas):
...
+    uof_citizens = pd.read_csv(deba.data("clean/uof_citizens_new_orleans_pd_2016_2021.csv"))
+    uof = pd.read_csv(deba.data("clean/uof_new_orleans_pd_2016_2021.csv"))
     post_event = pd.read_csv(deba.data("match/post_event_new_orleans_pd.csv"))
     award = pd.read_csv(deba.data("match/award_new_orleans_pd_2016_2021.csv"))
     lprr = pd.read_csv(deba.data("match/lprr_new_orleans_csc_2000_2016.csv"))
     sas = pd.read_csv(deba.data("match/sas_new_orleans_pd_2017_2021.csv"))
     complaints = fuse_cprr(cprr, actions, officer_number_dict)
-    use_of_force = fuse_use_of_force(uof, officer_number_dict)
-    personnel = fuse_personnel(pprr_ipm, lprr, pprr_csd, sas)
+    personnel = fuse_personnel(pprr_ipm, lprr, pprr_csd, sas, uof_officers)
```
pckhoi, line 161: Shouldn't you save the `uof_officers` in a file that begin with `fuse/uof_officers_` and end up in an actual table on wrglhub that is called `uof_officers`? Don't get me wrong, we should absolutely ensure that each `uid` in the `uof_officers` table should refer to a row in the `personnel` table of course. But the `uof_officers` should be a separate table. It's a many-to-one mapping between rows in `personnel` table and rows in `uof` table.

lib/clean.py
```diff
@@ -229,6 +229,17 @@ def clean_sexes(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
...
+        )
+        df = standardize_from_lookup_table(
+            df,
+            col,
+            [
+                [
+                    "male"
+                ],
+                ["female", "femaale", "famale"]
+            ]
```
pckhoi, line 242: Nice idea! I think all of the `str.replace` lines above can be integrated into this `standardize_from_lookup_table` call. Please add and/or modify tests though. We don't want to be over-confident about some changes and introduce bugs into the pipeline.

match/new_orleans_pd.py
```diff
@@ -210,15 +246,16 @@ def match_stop_and_search_to_pprr(sas, pprr_ipm):
     award = pd.read_csv(deba.data("clean/award_new_orleans_pd_2016_2021.csv"))
     lprr = pd.read_csv(deba.data("clean/lprr_new_orleans_csc_2000_2016.csv"))
     sas = pd.read_csv(deba.data("clean/sas_new_orleans_pd_2017_2021.csv"))
+    uof = pd.read_csv(deba.data("clean/uof_officers_new_orleans_pd_2016_2021.csv"))
```
pckhoi, line 249: perhaps call this variable `uof_officers` to avoid any confusion.

data/datavalid.yml
```diff
@@ -11,6 +11,8 @@ files:
     schema: stop_and_search
   fuse/appeals.csv:
     schema: appeal_hearing
+  fuse/citizen.csv:
+    schema: citizen
```
pckhoi, line 15: Like I said, please introduce `uof_officer` schema as well. this schema should be named `uof_citizen` instead of `citizen` because it is dependent on the use of force data.

.wrgl/config.yaml
```diff
@@ -1848,6 +1844,12 @@ branch:
     file: data/fuse/per_rayne_pd.csv
     primaryKey:
       - uid
+  citizen-new-orleans-pd:
+    remote: citizen-new-orleans-pd
+    merge: refs/heads/main
+    file: data/fuse/citizen_new_orleans_pd.csv
+    primaryKey:
+      - citizen_uid
```
pckhoi, line 1852: You should introduce new tables called `uof_citizens` and `uof_officers` here as well, and they are the merged results of data from multiple datasets, not just from New Orleans. It's a pain but we might as well make the changes now to make things consistent.

clean/new_orleans_pd_uof.py
ayyubibrahimi: I added all of the `use_of_force_*` columns to the `uof_officers` table. 

I also added the `division`, `division_level` and `unit` columns to the `uof_officers` table because they are data specific to each officer.

The `use_of_force_reason`, `division`, `division_level`, and `unit` do not need to be split, which is why I have written this function as I did. Is there a more efficient way?

fuse/new_orleans_pd.py
ayyubibrahimi: Ok.

match/new_orleans_pd.py
ayyubibrahimi: Got it.

lib/clean.py
ayyubibrahimi: Got it. I've left the code as is in this pull request with the idea that I will create a new branch for these changes.

clean/new_orleans_pd_uof.py
pckhoi: Sorry, I did not notice that `use_of_force_reason` does not need to be split. My feelings is to simply disregard the column meaning when it comes down to what columns should end up in `uof_officers`. Let's decide simply on whether the column is splittable or not (and they certainly shouldn't be citizen columns). The `tracking_number` and/or `uof_uid` are the only non-splittable columns that need to also be in `uof_officers`, for that's how we track relationships between `use_of_force` and `uof_officers`. I would argue that we only need `uof_uid` on the `uof_officers` table even though there's no harm to also have `tracking_number` in there.

So I would put `use_of_force_reason`, `division`, `division_level`, and `unit` back into the `use_of_force` table unless you have reasons to believe they might be different for each officer in another dataset. Sorry for not being clear here.

ayyubibrahimi: Got it! No worries at all. I went back and forth internally on this question. Glad it's cleared up. I'll make the changes.

ayyubibrahimi: Q. @pckhoi I initially generated a `uof_uid` for the `use_of_force` table because we are extracting the `occur_date` for the `uof_incident` event from the `use_of_force` table. Instead generating a `uof_uid` for the uof_officers table will leave these events without an identifier. Thoughts?

pckhoi: You can just generate `uof_uid` for the `use_of_force` table, then copy it to the `uof_officers` table. They have to be the same otherwise you cannot link the tables.

ayyubibrahimi: Got it. Is there a downside to also copying the `uof_uid` to the `citizens_table`? Or Is the `tracking_number` a sufficient link between the `use_of_force` table and the `uof_citizens` table?

pckhoi: Do you need `tracking_number` on the `uof_citizens` and `uof_officers`? If not then I recommend include only `uof_uid`. That's our id for each use of force record that we can guarantee to always be present and always have the same meaning no matter what the original dataset look like.

ayyubibrahimi: Makes perfect sense. I have only included `uof_uid`. Please review the latest changes at your convenience. Thanks!

ayyubibrahimi:
@pckhoi - When I run the commands you outlined, see below, the output is a csv in the new_orleans_pd directory. 

```
mkdir data/raw/new_orleans_pd
curl https://www.dropbox.com/s/7pzejv0xojp0os9/new_orleans_pd_uof_2016_2021.csv\?dl\=1 > data/raw/new_orleans_pd/new_orleans_pd_uof_2016_2021.csv
dvc add data/raw
dvc push
```

image

pckhoi:
Sorry, you need to give the `-L` flag to curl so that it follow redirect:
```
curl -L https://www.dropbox.com/s/7pzejv0xojp0os9/new_orleans_pd_uof_2016_2021.csv\?dl\=1 > data/raw/new_orleans_pd/new_orleans_pd_uof_2016_2021.csv
```

data/datavalid.yml
```diff
@@ -452,5 +400,46 @@ schemas:
...
+        description: >
+          unique identifier of use of force report.
+      - name: first_name
+      - name: last_name
+      - name: use_of_force_type
+      - name: use_of_force_level
+      - name: use_of_force_effective
+      - name: use_of_force_reason
+      - name: race
+      - name: sex
```
pckhoi, line 436: `first_name`, `last_name`, `race`, and `sex` never change so maybe we can just put them in the `personnel` table.

.wrgl/config.yaml
```diff
@@ -1848,6 +1848,18 @@ branch:
...
+    merge: refs/heads/main
+    file: data/fuse/uof_citizens_new_orleans_pd.csv
+    primaryKey:
+      - uof_citizen_uid
+  uof-officers-new-orleans-pd:
+    remote: uof-officers-new-orleans-pd
+    merge: refs/heads/main
+    file: data/fuse/uof_officers_new_orleans_pd.csv
+    primaryKey:
+      - uid
```
pckhoi, line 1862: Please add entries for `uof-citizen` and `uof-officer`.

data/datavalid.yml
ayyubibrahimi: Thanks.

.wrgl/config.yaml
ayyubibrahimi: Thanks. Just pushed the changes.

PR #259, opened 2022-03-01
Add TimedAccessor

pckhoi:
1. Added `lib.pandas_extensions.TimedAccessor`. This accessor allow you to time piped function execution:

```python
# just import this module for this accessor to become available
import lib.pandas_extensions

...

# replace df.pipe with df.timed.pipe will time my_func and print duration
df.timed.pipe(my_func, *args, **kwargs)
```

2. Upgrade `deba` to v0.0.6 which allow using `deba` outside of the root folder. For example inside a notebook:

```python
# deba.yaml is in the parent folder
import deba; deba.set_root('..')

...

# call deba.data like normal
df = pd.read_csv(deba.data('my_data.csv'))
```

### DIFF
--- a/lib/pandas_extensions.py
+++ b/lib/pandas_extensions.py
@@ -1,0 +1,40 @@
+import time
+
+import pandas as pd
+
+
+@pd.api.extensions.register_dataframe_accessor("timed")
+class TimedAccessor:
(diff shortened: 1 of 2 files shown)


ayyubibrahimi (approved):
Awesome.

PR #260, opened 2022-03-02
bugs/misc

ayyubibrahimi:
- extracted an `action` column and `disposition` column from the `allegation` column for cprr_st_tammany


pckhoi (approved):
Please use a more descriptive branch name and PR name next time. This PR seems to be very specific.

PR #269, opened 2022-03-22
extraction_post

ayyubibrahimi:
… reports

### DIFF
--- a/extraction_clean_spacy_output/clean_spacy_output_post.py
+++ b/extraction_clean_spacy_output/clean_spacy_output_post.py
@@ -1,0 +1,287 @@
+import pandas as pd
+import os
+import re
+from lib.uid import gen_uid
+import deba
+from lib.clean import clean_names
+from lib.columns import clean_column_names
+
+
+def clean_post_agency():
+    # split row if file contians multiuple accused names
+    directory = os.listdir(deba.data("raw/post/spacy"))
+    os.chdir(deba.data("raw/post/spacy"))
+    for hearing in directory:
+        if hearing.endswith(".csv"):
+            df = pd.read_csv(hearing).pipe(clean_column_names)
+
+            for columns in df.columns:
+                if "agency" in df.columns:
+                    data = (
+                        df.agency.astype(str)
+                        .str.lower()
+                        .str.strip()
+                        .str.replace(
+                            r"( ?= ?| ?_ ?| ?\' ?| ?bÃ¢â\‚ ?(¬â€Ã¢â\‚)? ?(¬Ëœ)? ?)",
+                            "",
+                            regex=True,
+                        )
+                        .str.extract(
+                            r"^(.+) (full-time|reserve|retired) ? ?(\w{1,2}\/\w{1,2}\/\w{4}) ? ?(\w{1,2}\/\w{1,2}\/\w{4})? ?(termination|voluntary (.+))?"
+                        )
+                    )
+                    df.loc[:, "agency"] = data[0]
+                    df.loc[:, "employment_status"] = data[1]
+                    df.loc[:, "hire_date"] = data[2]
+                    df.loc[:, "left_date"] = data[3]
+                    df.loc[:, "left_reason"] = data[4]
+                if "agency" not in df.columns:
+                    break
+                if "agency_1" in df.columns:
+                    data = (
+                        df.agency_1.astype(str)
+                        .str.lower()
+                        .str.strip()
+                        .str.replace(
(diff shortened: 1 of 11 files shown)


ayyubibrahimi:
@pckhoi - This is by no means close to finished. It is still very much in a theoretical stage as I have yet to figure out how to integrate it properly into this pipeline, despite my being able to run it locally. I wanted to get this in front of your eyes before our meeting tomorrow night with Tarak.

pckhoi (commented):
Here's the run dependency I picked up in pseudo-Makefile syntax:

```makefile
raw/post/download/*: raw/post/token/token.txt
	extraction_download/db_download_data.py

raw/post/metadata/post_metadata.csv: raw/post/token/token.txt
	extraction_download/db_download_metadata.py

raw/post/metadata/clean_post_metadata.csv: raw/post/metadata/post_metadata.csv
	extraction_download/db_clean_metadata.py

raw/post/ocr/*: raw/post/download/*
	extraction_ocr/ocr.py

raw/post/ocr/*: raw/post/ocr/*
	extraction_preprocess_text/preprocess.py

raw/post/json/padme.model raw/post/spacy/*: raw/post/ocr/* raw/post/json/ayyub.jsonl
	extraction_model/spacy_model.py

raw/post/spacy/*: raw/post/spacy/* raw/post/metadata/clean_post_metadata.csv
	extraction_download/db_append_metadata.py

raw/post/spacy/*: raw/post/spacy/*
	extraction_clean_spacy_output/clean_spacy_output_post.py

raw/post/spacy/*: raw/post/spacy/* raw/fused/personnel.csv
	extraction_match_spacy_output/extraction_match.py

raw/post/fused/padme.csv: raw/post/spacy/*
	extraction_fuse/extraction_fuse.py
```

There are some problems with this:

1. You probably want to run the download scripts on an ad-hoc basis (not run via make). Therefore you should not include them in the Deba pipeline. So please move the scripts under `extraction_download` to the `scripts` folder. Whenever there is a new file on Dropbox, run the scripts and `dvc add` the downloaded files.
2. `raw/post/token/token.txt` is a secret file, not a data file. We should store secrets in `.env` file and then read with `dotenv` package.
3. Some scripts have statements that are neither inside the main block nor a referenced function. Please keep in mind that Deba only read statements within the main block and any block reachable from the main block.
4. Many scripts read and write all files under a single directory, which means Deba can't pick up the prerequisites and targets from reading the script. For such scripts, you want to exclude them from `deba.yaml` and write Make rules for them. You can use the [wildcard](https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html) function to get file names under a directory and use the [patsubst](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html) function to transform them in any way you like. For example:

   ```makefile
   OCR_FILES := $(wildcard data/ocr/*.pdf)
   SPACY_FILES := $(patsubst data/ocr/%.pdf,data/spacy/%.csv,$(OCR_FILES))

   $(SPACY_FILES): $(OCR_FILES)
   	some_script.py
   ```

5. Some scripts even read from and write to the same file. This is the kind of hair-twisting script that even Make doesn't know how to run. So please avoid them at all costs. Ideally, a script should move data from one stage to another. For example, if it reads data from the `ocr` folder then it should output to another folder such as `ner`. If a script is too long, then you can split it into several functions, or even put the functions in a `lib` module.
6. Deba can't pick up names built on the fly like this:
   ```python
   file_name = file + ".csv"
   path = deba.data("raw/post/spacy/")
   df.to_csv(path + file_name, index=False)
   ```
   If you're reading/writing all files under a directory, then it doesn't matter. You can write Make rules as suggested in bullet 4. If you're only dealing with a few files and the file name does not need to be constructed dynamically then by all means, please don't.

With all that said I think your works can be organized into the following stages:

- `ocr`: all OCR tasks.
- `ner_metadata`: clean raw metadata file. I name this stage `ner_metadata` because the file `clean_post_metadata.csv` only goes into the `ner` stage. We could also just fold this stage into the `ner` stage to keep things simple. What do you think?
- `ner`: the `spacy_model.py` script can live here. I don't like the word `spacy` anywhere in the file path because it doesn't say anything about the purpose of the script. Imagine how bad it would be if we put `pandas` in the name of every script that uses Pandas.
- `clean`: clean `ner` output. We should not use another stage name such as `clean_post` unless there is a clear reason. The less stages there are, the easier it is to understand and maintain.
- `fuse`: fuse `clean` output.

ayyubibrahimi:
Thank you v much for all the feedback. Is there any reason that I should hold off with exploring the integration of this until Tarak's extraction pipeline has been formally integrated?

pckhoi:
Yes, I think that's the best way to go forward. Feel free to ask me to comment on anything and/or question in the mean time.

ayyubibrahimi:
Great. I'll do some studying on these concepts and circle back with comments and questions soon.

PR #272, opened 2022-03-29
bug/deduplicate

ayyubibrahimi:


clean/ipm_new_orleans_iapro_pprr.py
```diff
@@ -163,12 +170,15 @@ def clean():
         .pipe(clean_races, ["race"])
         .pipe(clean_department_desc)
         .pipe(assign_agency)
-        .pipe(gen_uid, ["agency", "employee_id"])
+        .pipe(
+            gen_uid, ["agency", "employee_id", "first_name", "last_name", "middle_name"]
+        )
```
pckhoi, line 175: Hmm, is the `employee_id` not unique enough for this dataset?

match/bossier_city_pd.py
```diff
@@ -1,9 +1,32 @@
-from datamatch import JaroWinklerSimilarity, ThresholdMatcher, ColumnsIndex
+from datamatch import JaroWinklerSimilarity, ThresholdMatcher, ColumnsIndex, Swap
```
pckhoi, line 1: Doesn't look like you're using `Swap` in this script.

match/cameron_so.py
```diff
@@ -2,6 +2,29 @@
...
+    df.loc[:, "fc"] = df.first_name.fillna("").map(lambda x: x[:1])
+    matcher = ThresholdMatcher(
+        ColumnsIndex("fc"),
+        {
+            "first_name": JaroWinklerSimilarity(),
+            "last_name": JaroWinklerSimilarity(),
+        },
+        df,
+    )
+    decision = 1
```
pckhoi, line 20: Deduplicate with threshold of `1` feels like a no-op to me. Unless I'm missing something.

match/new_orleans_pd.py
```diff
@@ -5,9 +5,61 @@
...
+
+    matcher = ThresholdMatcher(
+        ColumnsIndex("employee_id"),
+        {
+            "first_name": JaroWinklerSimilarity(),
+            "last_name": JaroWinklerSimilarity(),
+        },
+        df,
+        variator=Swap("first_name", "last_name"),
+    )
```
pckhoi, line 33: Are there people with the same `employee_id` but different names? I find this index choice rather odd.

match/new_orleans_pd.py
```diff
@@ -5,9 +5,61 @@
...
+    df.loc[:, "lc"] = df.last_name.fillna("").map(lambda x: x[:1])
+    matcher = ThresholdMatcher(
+        ColumnsIndex(["fc", "lc", "mc"]),
+        {
+            "first_name": JaroWinklerSimilarity(),
+            "last_name": JaroWinklerSimilarity(),
+        },
+        df,
+        variator=Swap("first_name", "last_name"),
+    )
```
pckhoi, line 56: `fc` is the first character of `first_name`. If `fc` is in the index, only rows with the same character for first name can match. The `Swap` variator is run after the index, therefore it will not have an effect in most cases. I.e. "John Smith" will never be able to match with "Smith John" because they belong to different indices, 'j' != 's'. Same thing with `lc`. If you still want to use the `Swap` variator then please avoid using `fc` and `lc` in the index. I guess you can still use `mc`, however in a dataset that is so inconsistent that the first name and last name order can vary then how consistent can the middle name be?

pckhoi:
Are all the match scripts that you added deduplicate step need it? It seems sudden that a lot of scripts receive this step.

ayyubibrahimi:
I will re-check my work. I ran the function on each match script because LLEAD is heavily populated with duplicates. I will ping you after I re-check for your review.

clean/ipm_new_orleans_iapro_pprr.py
ayyubibrahimi: I think that it is possible for two unique persons to have the same `employee_id` at two different points in time.

match/bossier_city_pd.py
ayyubibrahimi: Thank you.

match/cameron_so.py
ayyubibrahimi: Thanks.

match/new_orleans_pd.py
ayyubibrahimi: I haven't actually checked, but I know that it is very possible because of the timeframe for which we have data, therefore, I made this change to err on the side of caution. Would you prefer that I check first?

match/new_orleans_pd.py
ayyubibrahimi: Great point. Thank you

match/new_orleans_pd.py
pckhoi: No need to check first. I think it is fine if there's no instance where the same person has different employee ids.

match/new_orleans_pd.py
pckhoi: But then less meticulous agencies can also decide to not record the middle name, no? I'd say unless you have found instances where the name is written in different order then just avoid using the `Swap` variator. And if you do use it then please find some other columns that can be consistent. One idea I have is to sort `fc` and `lc` in alphabetical order and then concatenate them. That column will be way more resilient.

ayyubibrahimi: I've removed it. This is ready for review. Thanks!

pckhoi (approved):
LGTM

PR #279, opened 2022-04-07
enhancement/cross_agency

ayyubibrahimi:


ayyubibrahimi:
@pckhoi - I wanted to get some initial feedback before moving forward.

At the moment, matched clusters, produced by `cross_match_officers_between_agencies` are input into `create_person_table`. This should no longer be the case, does that sound correct? As the clusters produced by `cross_match_officers_between_agencies` are now being used for the purpose of requesting POST officer history reports from POST, rather than a valid input.

fuse/cross_agency.py
```diff
@@ -54,20 +54,18 @@ def assign_max_col(events: pd.DataFrame, per: pd.DataFrame, col: str):
 
 def read_constraints():
     # TODO: replace this line with the real constraints data
-    constraints = pd.DataFrame([], columns=["type", "uids"])
+    constraints = pd.read_csv(deba.data("clean/post_officer_history.csv"))
```
pckhoi, line 57: Are we really going to retire the contraints table? I think we still might want to manually edit the result from time to time. It's better to introduce POST history as an additional input.

clean/post_ohr.py
```diff
@@ -0,0 +1,52 @@
...
+    df.loc[:, "uids"] = df.uids.str.cat(df.matched_uid_7.fillna(""), sep=",")
+    df.loc[:, "uids"] = df.uids.str.cat(df.matched_uid_8.fillna(""), sep=",")
+
+    df.loc[:, "uids"] = df.uids.str.replace(r"\,\, ?(.+)?", "", regex=True)
+    return df.fillna("")
+
+
+def clean():
+    df = (
+        pd.read_csv(deba.data("raw/post/post_officer_history.csv"))
```
pckhoi, line 37: How did you produce this file? Seeing how it has numerous "matched_uid" columns I think it was clearly produced off of existing outputs. It's not a good idea to hide the initial clean, and match code. Once I add the OCR stage to this repo, you should commit the OCR code also.

fuse/cross_agency.py
ayyubibrahimi: Ok.

clean/post_ohr.py
ayyubibrahimi: Sounds great. Yes, I have a script for OCR, ner and extraction, clean, match, and fuse.

ayyubibrahimi:
@pckhoi I've added POST as a separate input. 

Additionally, after reviewing the cross_agency func tion, I was thinking that we might want to do the inverse re common_names, i.e., officers with uncommon names should receive a score of 1. I think that we can confidently match profiles of individuals with uncommon names without the need to cross-reference their data with POST. Thoughts?

clean/post_ohr.py
pckhoi: Please include the clean and match code for officer history in this PR.

fuse/cross_agency.py
```diff
@@ -172,9 +179,14 @@ def cross_match_officers_between_agencies(personnel, events, constraints):
...
 
-def create_person_table(clusters, personnel, personnel_event):
+def create_person_table(clusters, personnel, personnel_event, post):
     # add back unmatched officers into clusters list
-    matched_uids = frozenset().union(*[s for s in clusters])
+    matched_uids_post = frozenset().union(*[s for s in post["uids"]])
+    matched_uids_clusters = frozenset().union(*[s for s in clusters])
+
+    list_matched_uids = [matched_uids_post, matched_uids_clusters]
+    matched_uids = frozenset().union(*list_matched_uids)
```
pckhoi, line 188: Not sure how well this works but I prefer assigning a `history_id` that is unique per history record. And then add `AbsoluteScorer("history_id", 1, ignore_key_error=True),` to the `MaxScorer` which essentially make sure all uids that share the same history_id always match.

clean/post_ohr.py
ayyubibrahimi: Extracts `officer_name`, `officer_sex`, and the `agency` row which, once cleaned, is split into multiple columns: `agency`, `hire_date`, `left_date`, `left_reason`.

clean/post_ohr.py
```diff
@@ -0,0 +1,537 @@
...
+    clean_post_agency,
+)
+from lib.uid import gen_uid
+
+
+def clean_post_names(df):
+    names = (
+        df.officer_name.str.lower()
+        .str.strip()
+        .str.extract(r"(\w+\'?\w+?)\,? (\w+) ?(\w+)?")
```
pckhoi, line 15: I don't think `\w+?` has the meaning that you probably think it has here. `+?` is non-greedy one-or-more matching, it doesn't mean "any amount". So you might want to replace this pattern with
```python
r"(\w+(?:'\w+)?),? (\w+)(?: (\w+))?"
```

clean/post_ohr.py
```diff
@@ -0,0 +1,537 @@
...
+            r"?(\w{1,2}\/\w{1,2}\/\w{4}) ? ?(\w{1,2}\/\w{1,2}\/\w{4})? ?(termination|(\w+)? "
+            r"?(resignation)|(\w+)? ?(resigned)|retired)?"
+        )
+    )
+    df.loc[:, "agency_8"] = agency8[0]
+    df.loc[:, "employment_status_8"] = agency8[1]
+    df.loc[:, "hire_date_8"] = agency8[2]
+    df.loc[:, "left_date_8"] = agency8[3]
+    df.loc[:, "left_reason_8"] = agency8[4]
+    return df
```
pckhoi, line 477: This looks like a lot of copy-pasted code. Can you extract this out to functions? Remember DRY principle.

match/post_ohr.py
```diff
@@ -0,0 +1,333 @@
...
+    pprr = pd.read_csv(deba.data("raw/fused/event.csv"))
+    cprr = match_uid_and_personnel(cprr, pprr)
+    cprr = match_uid_1_and_personnel(cprr, pprr)
+    cprr = match_uid_2_and_personnel(cprr, pprr)
+    cprr = match_uid_3_and_personnel(cprr, pprr)
+    cprr = match_uid_4_and_personnel(cprr, pprr)
+    cprr = match_uid_5_and_personnel(cprr, pprr)
+    cprr = match_uid_6_and_personnel(cprr, pprr)
+    cprr = match_uid_7_and_personnel(cprr, pprr)
+    cprr = match_uid_8_and_personnel(cprr, pprr)
```
pckhoi, line 331: Are these functions all that different? I think you can parameterize and consolidate them.

fuse/cross_agency.py
```diff
@@ -147,6 +157,7 @@ def cross_match_officers_between_agencies(personnel, events, constraints):
                 ),
                 # but if two officers belong to the same attract constraint then give them the highest score regardless
                 AbsoluteScorer("attract_id", 1, ignore_key_error=True),
+                AbsoluteScorer("history_id", 1, ignore_key_error=True),
```
pckhoi, line 160: Where is this `history_id` column generated?

match/post_ohr.py
```diff
@@ -0,0 +1,333 @@
...
+        .str.replace(r"^\,", "", regex=True)
+    )
+    df = df.fillna("")
+    df = df[["post_id"]]
+    return df[~((df.post_id == ""))]
+
+
+if __name__ == "__main__":
+    cprr = pd.read_csv(deba.data("clean/post_officer_history.csv"))
+    pprr = pd.read_csv(deba.data("raw/fused/event.csv"))
```
pckhoi, line 322: Why is the `event.csv` called `pprr`? Is this the same file as `fuse/event.csv`?

match/post_ohr.py
```diff
@@ -0,0 +1,333 @@
...
+    df.loc[:, "post_id"] = df.post_id.str.cat(df.matched_uid_5.fillna(""), sep=",")
+    df.loc[:, "post_id"] = df.post_id.str.cat(df.matched_uid_6.fillna(""), sep=",")
+    df.loc[:, "post_id"] = df.post_id.str.cat(df.matched_uid_7.fillna(""), sep=",")
+    df.loc[:, "post_id"] = df.post_id.str.cat(df.matched_uid_8.fillna(""), sep=",")
+
+    df.loc[:, "post_id"] = (
+        df.post_id.fillna("")
+        .str.replace(r"\,\, ?(.+)?", "", regex=True)
+        .str.replace(r"^\,", "", regex=True)
+    )
```
pckhoi, line 314: You're overusing regex. Regex can be very slow so use them sparingly. For the purpose of joining columns, just use `DataFrame.apply`:
```python
df.loc[:, 'post_id'] = df[
    [col for col in df.columns if col.startswith('matched_uid_')]
].apply(lambda sr: ','.join(sr[sr.notna()].to_list()))
```
```

File: clean/post_ohr.py (hunk 4 above)
Comment (ayyubibrahimi) [reply clean/post_ohr.py:15]: Thanks

File: clean/post_ohr.py (hunk 5 above)
Comment (ayyubibrahimi) [reply clean/post_ohr.py:477]: Thank you for the reminder!

File: match/post_ohr.py (hunk 8 above)
Comment (ayyubibrahimi) [reply match/post_ohr.py:322]: It's poorly named. It's `fuse/event.csv` and `fuse/personnel.csv` merged into one table to match with `first_name`, `last_name`, and `agency`. I will correct.

File: fuse/cross_agency.py (hunk 7 above)
Comment (ayyubibrahimi) [reply fuse/cross_agency.py:160]: Typo. I will correct to `post_id`, which is generated in `match/post_ohr.py` as it is the field in which matched uids are concatenated. Do you think that `history_id` is a more appropriate name than `post_id`?

File: match/post_ohr.py (hunk 9 above)
Comment (ayyubibrahimi) [reply match/post_ohr.py:314]: Thanks

File: fuse/cross_agency.py (hunk 7 above)
Comment (pckhoi) [reply fuse/cross_agency.py:160]: I don't think the `post_id` is what I had in mind. In fact, I don't see why the uids need to be concatenated. The `history_id` could just be the frame index at the point where you have multiple uid columns. Then instead of concatenating uids, stack the frame instead. That way you will end up with 2 columns: `history_id` and `uid`. I hope that makes sense.

Also you need to make sure there are no uid that ends up in 2 different history records. It is quite easy to check.

File: match/post_ohr.py (hunk 8 above)
Comment (pckhoi) [reply match/post_ohr.py:322]: Okay, again don't put a file that was produced off of outputs of this pipeline into the raw folder. You're essentially cutting this file off of the pipeline so it will never be updated when either `fuse/event.csv` or `fuse/personnel.csv` is updated. And how this file came to be is also totally unknown. How can other people update it afterward?

Comment (ayyubibrahimi) [reply match/post_ohr.py:322]: What is the best way to match these officers with those officers who exist in `fuse/personnel.csv`

Comment (pckhoi) [reply match/post_ohr.py:322]: Simplest way: join the `first_name`, `last_name`, `middle_name` from `personnel` into `event` using `uid`. Then match using `ColumnsIndex(['fc', 'agency'])` and JaroWinkler on `first_name`, `last_name`, `middle_name`. Pretty straight-forward I think. Unless you have other columns that can also be used to match.

Comment (ayyubibrahimi) [reply match/post_ohr.py:322]: On it. Thank you.

Comment (ayyubibrahimi) [reply match/post_ohr.py:322]: Sorry for the misunderstanding. I think that I am still misunderstanding.

I was under the impression that at the `match` stage we should not be calling tables produced at the `fuse` stage, namely `fuse/personnel.csv` and `fuse/event.csv` because it would mess with the dependencies.

Comment (pckhoi) [reply match/post_ohr.py:322]: You're right, It is such a bind but for now just make the new script a `fuse` script. Make all the scripts that depend on it a `fuse` script also. we can redistribute the scripts into existing stages or create new stages once we have the complete picture. Feel free to suggest new stages. I don't think `clean`, `match`, and `fuse` are sufficient anymore.

Comment (pckhoi) [reply match/post_ohr.py:322]: After thinking about it some more. I think it's best to give Deba an option to disable stage order enforcing and instead make sure that Deba can detect circular dependency. That way you can have `match` script read from `fuse` output and vice versa but the whole pipeline would still form a DAG.

Comment (pckhoi) [reply match/post_ohr.py:322]: Looks like Make already drop circular dependency by itself so all the changes I need to make is to add an option that disable stage order enforcement.

Comment (ayyubibrahimi) [reply match/post_ohr.py:322]: Much better idea than what I was going to suggest. Awesome. I've been traveling a lot in the past few days, but I hope to have my changes to the script committed soon as well.

Comment (ayyubibrahimi) [reply match/post_ohr.py]: I just pushed what I was going to suggest just for your general feedback. It is irrelevant now that you've decided the best solution, however, I wanted your feedback because while I think that it's not a great solution I can't exactly describe why. Would love your feedback. 

Still in the process of parameterizing and consolidating the match.

File: lib/clean.py [hunk 10]
```diff
@@ -790,3 +790,152 @@ def convert_dates(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
...
+            .str.replace(r"^Harbor PD$", "New Orleans Harbor PD", regex=True)
+            .str.replace(r"^Lafayette$", "Lafayette PD", regex=True)
+            .str.replace(r"^Louisiana State$", "Louisiana State PD", regex=True)
+            .str.replace(r"^Caddo Parish$", "Caddo SO", regex=True)
+        )
+
+    return df
+
+
+def clean_post_agency_row(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
```
pckhoi, line 887: The `clean_post_agency_*` functions do not feel like they belong in `lib/clean.py` because they are not reused outside of `post_history.py`.

fuse/post_history.py
```diff
@@ -0,0 +1,310 @@
...
+    return dfb
+
+
+def match_uid(dfb, post, dfa=None):
+    matcher = ThresholdMatcher(
+        ColumnsIndex(["fc"]),
+        {
+            "first_name": JaroWinklerSimilarity(),
+            "last_name": JaroWinklerSimilarity(),
+            "agency": JaroWinklerSimilarity(),
```
pckhoi, line 217: If the `agency` column can be exactly matched (they should be), then just put it in the `ColumnsIndex`.

fuse/post_history.py
```diff
@@ -0,0 +1,310 @@
...
+        .pipe(gen_uid, ["first_name", "last_name", "middle_name", "agency_3"], "uid_3")
+        .pipe(gen_uid, ["first_name", "last_name", "middle_name", "agency_4"], "uid_4")
+        .pipe(gen_uid, ["first_name", "last_name", "middle_name", "agency_5"], "uid_5")
+        .pipe(gen_uid, ["first_name", "last_name", "middle_name", "agency_6"], "uid_6")
+        .pipe(gen_uid, ["first_name", "last_name", "middle_name", "agency_7"], "uid_7")
+        .pipe(gen_uid, ["first_name", "last_name", "middle_name", "agency_8"], "uid_8")
+        .pipe(drop_rows_missing_agency)
+        .pipe(drop_rows_missing_names)
+    )
+    return df
```
pckhoi, line 130: There's a cleaner way to extract and present uids. Look at this example:
```python
>>> df = pd.DataFrame([['john', 'doe', 'm', 'brusly pd', 'caddo pd'], ['jane', 'lane', 'c', 'new orleans pd', 'caddo pd']], columns=['first_name', 'last_name', 'middle_name', 'agency_1', 'agency_2'])
>>> df
  first_name last_name middle_name        agency_1  agency_2
0       john       doe           m       brusly pd  caddo pd
1       jane      lane           c  new orleans pd  caddo pd
>>> stacked_agency_sr = df[['agency_1', 'agency_2']].stack()
>>> stacked_agency_sr
0  agency_1         brusly pd
   agency_2          caddo pd
1  agency_1    new orleans pd
   agency_2          caddo pd
dtype: object
>>> stacked_agency_df = stacked_agency_sr.reset_index().iloc[:, [0, 2]]
>>> stacked_agency_df.columns = ['history_id', 'agency']
>>> stacked_agency_df
   history_id          agency
0           0       brusly pd
1           0        caddo pd
2           1  new orleans pd
3           1        caddo pd
>>> names_df = df[['first_name', 'last_name', 'middle_name']].reset_index()
>>> names_df = names_df.rename(columns={'index': 'history_id'})
>>> names_df
   history_id first_name last_name middle_name
0           0       john       doe           m
1           1       jane      lane           c
>>> stacked_agency_df.merge(names_df, on='history_id', how='outer')
   history_id          agency first_name last_name middle_name
0           0       brusly pd       john       doe           m
1           0        caddo pd       john       doe           m
2           1  new orleans pd       jane      lane           c
3           1        caddo pd       jane      lane           c
```

Then just generate uid like normal. This way you only have to deal with one `agency` and one `uid` column.

match/post_ohr.py
pckhoi: @ayyubibrahimi Once Deba no longer enforce stage order, you should just put matching code in match scripts, cleaning code in clean scripts. etc. Putting all kinds of code in a single script also work but there are no structure, over time the project slowly descend into chaos. There always have to be some kind of structure in a project so that A) you know instantly where to put a piece of code and B) it becomes predictable and easy for others to maintain your code.

Since we cannot really structure this project order-wise, we have to structure it function-wise: organizing scripts based on their purpose. For now they will be `clean`, `match`, and `fuse` but in the future we will add `ocr` and `ner` as well.

fuse/post_history.py
ayyubibrahimi: Thank you v much.

match/post_ohr.py
ayyubibrahimi: Makes perfect sense. Thank you.

pckhoi: @ayyubibrahimi #283 should do it

ayyubibrahimi: Awesome. Thanks for that.

ayyubibrahimi:
Hey Khoi,

Just merged main into this branch and I receive the following error:

```
make: [deba.mk:33: /clean/baton_rouge_csd_pprr.py.md5] Error 1 (ignored)
md5sum clean/baton_rouge_csd_pprr.py > /clean/baton_rouge_csd_pprr.py.md5
/bin/bash: /clean/baton_rouge_csd_pprr.py.md5: No such file or directory
make: *** [deba.mk:34: /clean/baton_rouge_csd_pprr.py.md5] Error 1
```

Is this another issue with my line endings?

pckhoi:
Please install the latest deba with `pip install -r requirements.txt`, then run `python -m deba md5Dir` inside the repo and let me know what you see.

ayyubibrahimi:
It works. Thanks.

clean/post_officer_history.py
```diff
@@ -0,0 +1,255 @@
...
+
+def extract_agency(df):
+    for col in df.columns:
+        if col.startswith("agency"):
+            df[col] = (
+                df[col]
+                .fillna("")
+                .str.lower()
+                .str.strip()
+                .str.replace(r"(\w{1}? ?\w+ ?\w+? ?\w+) (.+)", r"\1", regex=True)
```
pckhoi, line 82: `{1}` is completely redundant. `\w` always matches exactly one letter .

clean/post_officer_history.py
```diff
@@ -0,0 +1,255 @@
...
+                "agency_3",
+                "agency_4",
+                "agency_5",
+                "agency_6",
+                "agency_7",
+                "agency_8",
+            ],
+        )
+        .pipe(clean_post_agency_column)
+        .pipe(drop_rows_with_only_one_agency)
```
pckhoi, line 244: I think there's a possibility of using records with just one agency to filter out false positives in cross agency matching results.

clean/post_officer_history.py
```diff
@@ -0,0 +1,255 @@
...
+        .pipe(drop_rows_missing_names)
+        .pipe(generate_history_id)
+        .pipe(gen_uid, ["first_name", "middle_name", "last_name", "agency"])
```
pckhoi, line 247: Are there no employee id that come with officer history data?

lib/clean.py
pckhoi: Please remove redundant functions from this file.

match/post_officer_history.py
```diff
@@ -0,0 +1,42 @@
...
+            "middle_name": JaroWinklerSimilarity(),
+        },
+        dfa,
+        dfb,
+    )
+    decision = 0.90
+    matches = matcher.get_index_pairs_within_thresholds(decision)
+    match_dict = dict(matches)
+
+    post.loc[:, "matched_uid"] = post.uid.map(lambda x: match_dict.get(x))
```
pckhoi, line 33: I think you can just assign to `post.uid` here. We don't have any use of unmatched uid anyhow.

fuse/join_events_personnel_tables.py
```diff
@@ -0,0 +1,20 @@
...
+        dfb[["first_name", "last_name", "middle_name", "uid"]],
+        on="uid",
+        how="outer",
+    )
+    return df.dropna().drop_duplicates(subset=["uid"])
+
+
+if __name__ == "__main__":
+    df = join_event_and_personnel()
+    df.to_csv(deba.data("fuse/events_and_personnel_merged.csv"), index=False)
```
pckhoi, line 20: This file seems redundant.

clean/post_officer_history.py
ayyubibrahimi: Thanks

clean/post_officer_history.py
ayyubibrahimi: Unfortunately not

fuse/join_events_personnel_tables.py
ayyubibrahimi: Should this function be moved to the match script where I match post and events/personnel?

clean/post_officer_history.py
ayyubibrahimi: I don't think that I follow.

pckhoi: If we know that John Smith from Baton Rouge never worked at anywhere else. And then we see that the cross agency script suggest that there's a John Smith from Caddo PD who is a match then we can say that it is a false negative?

pckhoi: Don't worry about it if it sounds complicated.

fuse/join_events_personnel_tables.py
pckhoi: I said it is redundant because I don't see the output "fuse/events_and_personnel_merged.csv" used anywhere.

pckhoi:
@ayyubibrahimi Please merge my latest changes from `main` which ensure transaction is applied when PR is merged. And please make sure the process-data job can complete successfully of course.

fuse/join_events_personnel_tables.py
ayyubibrahimi: Thanks.

clean/post_officer_history.py
ayyubibrahimi: Makes perfect sense. On it.

ayyubibrahimi:
@pckhoi The process-data job has failed after the latest changed were merged from `main`. Please advise.

pckhoi:
I saw this error in "process-data" step:
```
running fuse/cross_agency.py build/person.csv
    stty: 'standard input': Inappropriate ioctl for device
    sys:1: DtypeWarning: Columns (3,5,8,9,10,12,13,14,15,16,17,18,19,20,21,22,23,24,[25](https://github.com/ppact/processing/runs/6226428697?check_suite_focus=true#step:10:25),28,29,30,31) have mixed types.Specify dtype option on import or set low_memory=False.
    Traceback (most recent call last):
      File "/runner/_work/processing/processing/fuse/cross_agency.py", line 303, in <module>
        post = read_post()
      File "/runner/_work/processing/processing/fuse/cross_agency.py", line 71, in read_post
        post = pd.read_csv(deba.data("match/post_officer_history.csv"))
    read personnel file (68389 x 9)
    read events file (1042204 x [32](https://github.com/ppact/processing/runs/6226428697?check_suite_focus=true#step:10:32))
    read constraints (0 rows)
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/io/parsers.py", line 610, in read_csv
        return _read(filepath_or_buffer, kwds)
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/io/parsers.py", line 462, in _read
        parser = TextFileReader(filepath_or_buffer, **kwds)
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/io/parsers.py", line 819, in __init__
        self._engine = self._make_engine(self.engine)
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/io/parsers.py", line 1050, in _make_engine
        return mapping[engine](self.f, **self.options)  # type: ignore[call-arg]
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/io/parsers.py", line 1867, in __init__
        self._open_handles(src, kwds)
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/io/parsers.py", line 1[36](https://github.com/ppact/processing/runs/6226428697?check_suite_focus=true#step:10:36)2, in _open_handles
        self.handles = get_handle(
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/io/common.py", line 6[42](https://github.com/ppact/processing/runs/6226428697?check_suite_focus=true#step:10:42), in get_handle
        handle = open(
    FileNotFoundError: [Errno 2] No such file or directory: '/runner/_work/data/data/match/post_officer_history.csv'
make: *** [wrgl.mk:12: /runner/_work/data/data/fuse/person.csv] Error 1
Error: Process completed with exit code 2.
```

I also have this error on local and it seems that not enforcing stage order caused this error. I made an edit to the workflow file that force update the deps files and committed to this branch.

pckhoi:
@ayyubibrahimi so it didn't work because there was a bug with Deba where it try to write rules for scripts that don't have any target. Upgrading Deba fixes the issue.

ayyubibrahimi:
Ok, great. I'll update now. Thanks!

pckhoi:
Please merge from `main` and push again.

ayyubibrahimi:
Cleared!

clean/post_officer_history.py
ayyubibrahimi: @pckhoi Is this what you had in mind re false negatives? Wanted to get your feedback on whether it should be in the form of a `false_neg_id` as in input to the cross_agency script.

pckhoi: Sorry, false positive, not false negative. But let me take a look.

clean/post_officer_history.py
```diff
@@ -0,0 +1,272 @@
...
+            raise ValueError("uid found in multiple history ids")
+
+    return df[~((df.agency.fillna("") == ""))]
+
+
+def generate_false_neg_id(df):
+    false_negs = [
+        ids == False for ids in df.duplicated(subset=["history_id"], keep=False)
+    ]
+    df.loc[:, "false_negatives"] = false_negs
```
pckhoi, line 220: Don't convert a series into a list and even discard elements from it. `df.duplicated(subset=["history_id"], keep=False)` is a boolean series which have the same index and the same length as `df`, so it is perfect to assign to a column already. `false_negs` not only  does not have any index (because it was converted into a list), it doesn't even have the same length. So I'm very curious how this assignment is even possible.

clean/post_officer_history.py
```diff
@@ -0,0 +1,272 @@
...
+    df.loc[:, "false_negatives"] = df.astype(str).false_negatives.str.replace(
+        "False", "", regex=False
+    )
+
+    false_df = df.pipe(
+        gen_uid, ["false_negatives", "history_id"], "false_neg_id"
+    ).drop_duplicates(subset=["false_neg_id"], keep=False)
+    false_df = false_df[["false_neg_id", "history_id"]]
```
pckhoi, line 228: The column should be named "switched_job" or something similar. I think this name is more descriptive and open this column up to more purposes. And you don't need to generate any id. Just keep it a boolean series:
```python
df.loc[:, 'switched_job'] = df.duplicated(subset=["history_id"], keep=False)
```

In the `cross_agency` script you are using `DissimilarFilter("false_neg_id")`, which attempts to filter out pairs that have the same `history_id` (`false_negatives` was also generated from `history_id` so in truth it all comes down to `history_id`). Doing this is not necessary because the records that we want to eliminate from the matching process all have only one `uid` associated with a `history_id`. Which mean it is only one row, not enough to form a pair. Not only that, I think it might wipe out true positives because rows with the same `history_id` (officers who switched job) will still have the same `false_neg_id`.

We don't need to filter at the pair level, we can simply filter at the record level. Add this line before cross_agency.py line 131 would be enough:
```python
per = discard_rows(per, per.switched_job, 'officers who has not switched job')
```

clean/post_officer_history.py
ayyubibrahimi: Thanks.

clean/post_officer_history.py
ayyubibrahimi: Makes sense. Thanks.

ayyubibrahimi:
@pckhoi please review when you're free. I removed `per = discard_rows(per, per.min_date.notna(), "officers with no event")` because only officers from `post_officer_history` were filtered out. That said, the reports do contain a `hire_date` and sometimes a `left_date`. Should these data be incorporated?

pckhoi:
I think the min_date filter cannot be removed because event timing is the backbone of the cross agency script. It sounds to me like the majority of records from post officer history cannot be matched to existing uids. Is that correct? The main question is: do we want to treat officer history records as regular personnel records or not? We have been treating officer history records as supplement data to support cross agency matching. And if we want to keep doing that, we should just discard any uid from officer history that couldn't be matched. If we want to import officer history records into personnel data however, we need to rethink our current approach.

ayyubibrahimi:
That is correct.

Agreed. 

My intuition is that those officers that can be matched to existing uids should be treated as supplementary data to support cross agency matching, and those officers who can't be matched should become personnel records that always receive a score of `1` in the `cross_agency` match script. What do you think?

pckhoi:
I take your answer as option 2 then. We will treat history records as regular personnel records. That means each record should produce a distinct `uid` and a set of hire/left events associated with it. The cross agency script already know how to deal with non-matches by merging them from `personnel` in the `create_person_table` function. So we can simply discard non-matches from POST officer history, as long as records are found in `personnel`.

Let's do it this way:
- Clean POST officer history
- Rename `fuse/personnel.csv` and `fuse/events.csv` output of `fuse/all.py` to `fuse/personnel_pre_post.csv` (keep `agency` column in this table for later matching) and `fuse/events_pre_post.csv`.
- Add a script `match/post_officer_history.py` which matches officer history with `fuse/personnel_pre_post.csv`. This script should produces `match/post_history.csv`, which has all the final uids.
- Add another script `fuse/post_officer_history.py` which extracts events and personnel out of `match/post_history.csv`, merges with `fuse/personnel_pre_post.csv` and `fuse/events_pre_post.csv`, and produces the final `fuse/personnel.csv` and `fuse/events.csv`.
- `fuse/cross_agency.py` can then process `fuse/personnel.csv` and `fuse/events.csv` as usual. It should read `match/post_history.csv` as authoritative data source, match the uids with the same history_id and discard non-matches as we discussed.

pckhoi:
Regarding fixing error in process-data job, try to reproduce the error on your local. It should run identical to your local. The only difference is it always start from a clean slate with the repo freshly checked out, dvc pulled, and wrgl pulled before Make update everything. All outputs from previous runs are still kept to speed up processing of course. But whether Make/Deba process data wrongly or not, they should behave identical to local.

One way to avoid errors and/or minimize differences compared to GitHub job is to always run `pip install -r requirements.txt` after you pull new changes from main.

ayyubibrahimi:
Sounds good. I'm on it.

ayyubibrahimi:
Re the latest failure of process-data (please see below): I'm confused as `agency` is in the personnel table on local. 

```
running match/post_officer_history.py
    Traceback (most recent call last):
      File "/runner/_work/processing/processing/match/post_officer_history.py", line 40, in <module>
        post = match_post_to_personnel(post, personnel)
      File "/runner/_work/processing/processing/match/post_officer_history.py", line 15, in match_post_to_personnel
        dfb = personnel[["uid", "first_name", "last_name", "middle_name", "agency"]]
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/core/frame.py", line 3030, in __getitem__
        indexer = self.loc._get_listlike_indexer(key, axis=1, raise_missing=True)[1]
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/core/indexing.py", line 1[26](https://github.com/ppact/processing/runs/6278349541?check_suite_focus=true#step:10:26)6, in _get_listlike_indexer
        self._validate_read_indexer(keyarr, indexer, axis, raise_missing=raise_missing)
      File "/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pandas/core/indexing.py", line 1[31](https://github.com/ppact/processing/runs/6278349541?check_suite_focus=true#step:10:31)6, in _validate_read_indexer
        raise KeyError(f"{not_found} not in index")
    KeyError: "['agency'] not in index"
```

pckhoi:
Just a side note but the New Orleans scripts are among the slowest. One of us should eventually try to profile them.

fuse/all.py
```diff
@@ -322,4 +322,4 @@ def fuse_brady():
     missing_agency_df = find_event_agency_if_missing_from_post(event_df, post_event_df)
     post_event_df = post_event_df[~post_event_df["agency"].isin(event_df["agency"])]
     event_df = pd.concat([event_df, post_event_df], ignore_index=True)
-    event_df.to_csv(deba.data("fuse/event.csv"), index=False)
+    event_df.to_csv(deba.data("fuse/events_pre_post.csv"), index=False)
```
pckhoi, line 325: Looks like typo here. Should be `fuse/event_pre_post.csv`. How was you able to run without issue on local?

ayyubibrahimi: Thanks. I went back and forth between naming the table `events_pre_post` and `event_pre_post` so I have both on my local.

fuse/cross_agency.py
```diff
@@ -104,6 +110,19 @@ def cross_match_officers_between_agencies(personnel, events, constraints):
     per = discard_rows(
         per, per.agency != "", "officers not linked to any event", reset_index=True
     )
+
+    per = pd.merge(
+        per,
+        post,
+        on=["uid", "first_name", "last_name", "agency"],
+        how="outer",
+    )
```
pckhoi, line 119: Do we need to include `first_name` and `last_name` when `uid` is already present? `post` probably contains multiple rows with the same `uid`. I wonder what effects would that have on the output. It's safer to deduplicate `post` on `uid` first.

match/post_officer_history.py
```diff
@@ -0,0 +1,41 @@
...
+    match_dict = dict(matches)
+
+    post.loc[:, "uid"] = post.uid.map(lambda x: match_dict.get(x, x))
+    return post
+
+
+if __name__ == "__main__":
+    post = pd.read_csv(deba.data("clean/post_officer_history.csv"))
+    personnel = pd.read_csv(deba.data("fuse/personnel_pre_post.csv"))
+    post = match_post_to_personnel(post, personnel)
```
pckhoi, line 40: We should check that `post` agencies are a subset of `personnel` agencies before matching.

fuse/cross_agency.py
ayyubibrahimi: Got it. Thanks.

match/post_officer_history.py
ayyubibrahimi: Thanks.

ayyubibrahimi:
Ready for review @pckhoi

ayyubibrahimi:
Made edits after process-data failed. It's now ready for review.

pckhoi (approved):
LGTM

PR #282, opened 2022-04-11
Enhancement/gh k8s runner

pckhoi:
- Add new workflows that facilitate continuous data integration.
- Remove `<some_func>().to_csv()` patterns which are not currently picked up by Deba. Deba might be able to pick them up with the right pattern though.
- Upgrade deba version
- Add Github k8s runner manifests and documentation


ayyubibrahimi (approved):
Awesome.

PR #288, opened 2022-05-02
Bugs/use of force

ayyubibrahimi:


pckhoi:
So strange, looking at the transaction I don't see the `use-of-force` table change. Did you see the change on local?

ayyubibrahimi:
Hmm. Yes. Changes are present on local.

pckhoi:
I added `datavalid.yml` as common prerequisite and it seems to trigger re-run correctly now. However there is an error with wrgl pull so hold on.

pckhoi:
@ayyubibrahimi there are only 2% of the uids populated in use_of_force. Please investigate.

ayyubibrahimi:
I see the issue. We decided to split out the `fuse/use_of_force.csv` table into three: `fuse/use_of_force.csv`, `fuse/uof_officers.csv`, `fuse/uof_citizens.csv`. 

Only `fuse/uof_officers.csv` contains `uid`, but these three tables can be joined on `uof_uid`.

 Kenner PD, the only other agency with a uof table, has not yet been split. Once Kenner PD is split, all uids will live in the `fuse/uof_officers.csv` table. I will do it now.

ayyubibrahimi:
@pckhoi This is ready for review.

pckhoi:
Hmm, why does the `use-of-force-officers` table have `uof_uid` as the primary key? There can be multiple `uid` per `uof_uid` so I think `use-of-force-officers` should have both `uof_uid` and `uid` as the primary key.

ayyubibrahimi:
Agreed. Thanks.

pckhoi (approved):
LGTM

PR #291, opened 2022-05-11
bugs/use_of_force

ayyubibrahimi:
- added `citizen_sex`, `citizen_race`, and `uof_uid` to `uof_citizen_uid` generation


pckhoi:
@ayyubibrahimi can you deduplicate the uof officers on `uid` and `uof_uid`, and then deduplicate uof citizens on `uof_citizen_uid` please? Not that it is still possible to commit with duplicated primary key but it is always better to control which row make it in the script.

ayyubibrahimi:
Makes sense! Please review

pckhoi (approved):
LGTM

PR #313, opened 2022-06-06
Fix circular dependencies for brady lists matching

pckhoi:
- In brady list matching scripts, Replace `fuse/personnel.csv` with other files (POST data in some cases) to eliminate circular dependencies.
- `pull_person` rule will only pull the `person` branch and nothing else.

@ayyubibrahimi  after this is merged, you should be able to run `make` and end up with the `.wrgl` folder just taking 125Mb which is just the `person` branch. I hope your computer can take on that much at least.

Also please review how I have changed the brady matching scripts. We really can't feed the brady output into generating `fuse/personnel.csv` only to feed this file into generating the brady files again. Just match the brady lists with any relevant personnel data from the same area. Or if there's nothing like that then just match against the POST data.

### DIFF
--- a/.wrgl/config.yaml
+++ b/.wrgl/config.yaml
@@ -2,7 +2,7 @@ remote:
   origin:
     url: https://hub.wrgl.co/api/users/ipno/repos/data
     fetch:
-      - +refs/heads/*:refs/remotes/data/*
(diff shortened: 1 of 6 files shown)


ayyubibrahimi (approved):
Awesome. Will do.

PR #314, opened 2022-06-09
bug/brady_lists

ayyubibrahimi:


match/baton_rouge_da.py
```diff
@@ -38,7 +38,7 @@ def match_brady_to_personnel(brady, per):
 
 
 if __name__ == "__main__":
-    pprr = pd.read_csv(deba.data("match/pprr_baton_rouge_pd_2021.csv"))
+    post = pd.read_csv(deba.data("clean/pprr_post_2020_11_06.csv"))
     brady = pd.read_csv(deba.data("clean/brady_baton_rouge_da_2021.csv"))
-    brady = match_brady_to_personnel(brady, pprr)
+    brady = match_brady_to_post(brady, post)
```
pckhoi, line 43: Why do you match to the entire post roster instead of just for Baton Rouge?

ayyubibrahimi: Placement on a brady list, implicitly, implies that the officer is no longer with the agency. It's more likely, I think, that the officer is in POST or the personnel table (a different agency) than in Baton Rouge because he/she was most likely fired after placement on the brady list.

pckhoi: I see where you are coming from. However this approach does have a problem. What if there are one or more people with the same name from other agencies? How do you tell them apart? You're basically attempting to match people cross-agency with nothing but names. Sure it might work but it also may produce lots of false positives.

Once somebody has moved on to other agencies, I think we should only rely on the cross agency script to discover where they end up.

And even if you just match the Brady list with roster for that agency, you might still find matches if the roster predate the Brady list. And even if there are no matches, it is fine, the record is still there, we can still try to find out where they end up in the cross agency script.

ayyubibrahimi: Makes sense. Made the changes. This is ready for review.

pckhoi (approved):
LGTM

PR #316, opened 2022-06-09
Enhancement/nopd seps

ayyubibrahimi:


datavalid.yml
```diff
@@ -60,6 +60,10 @@ schemas:
           - female
           - male
       - name: agency
+      - name: class_id
+      - name: assignment_id
```
pckhoi, line 64: Would be nice to add `description` to more obscured name like `class_id` and `assignment_id`.

ayyubibrahimi: Agreed. I have no idea what they actually mean, yet. Working on getting more information.

pckhoi: I think we should always discard columns that we don't understand. Data without meaning cannot serve any purpose. If the intention is to remind ourselves to find out more about some columns, then I think it's better to leave `# TODO:` comments in the clean scripts. Unless you have other reasons to leave them in.

ayyubibrahimi: My original thinking was that someone who is downloading these tables for research purposes may know what they mean, as they are likely to have more insight than us into NOPD. Happy to leave `# TODO:` comments in the clean scripts if you think that concern is unwarranted.

pckhoi: Yeah, it's just hard for me to imagine a case where someone who look at the data we produce and would know more than us. They will probably have to look at our scripts anyway and even the source data as well. And the web developers will have to modify their SQL schema to accommodate any new field. So it's probably more trouble than it's worth.

ayyubibrahimi: Makes sense to me. Removed the columns. This is ready for review.

pckhoi:
I see some changes that should belong to the ocr branch. Does this branch share some history? I'll review again after OCR changes are merged.

ayyubibrahimi:
@pckhoi Ready for review

dvc/data/raw/ner.dvc
```diff
@@ -0,0 +1,6 @@
+wdir: ../../../data/raw
+outs:
+- md5: 43d5c9ca303e4eccba73044b10eef279.dir
+  size: 262617
+  nfiles: 1
+  path: ner
```
pckhoi, line 6: This file seems redundant. Please remove it.

dvc/data/raw/post_officer_history.dvc
```diff
@@ -0,0 +1,6 @@
+wdir: ../../../data/raw
+outs:
+- md5: b8bc7e6f1d623ce5f582a781dcabded2.dir
+  size: 256033
+  nfiles: 1
+  path: post_officer_history
```
pckhoi, line 6: This file also seems redundant.

datavalid.yml
```diff
@@ -60,6 +60,8 @@ schemas:
           - female
           - male
       - name: agency
+      - name: years_of_service
+      - name: rank_desc
```
pckhoi, line 64: I think you should put `years_of_service` in the `event` table instead. `rank_desc` might change for an officer so it should not be added to `personnel`.

datavalid.yml
```diff
@@ -301,6 +303,7 @@ schemas:
         description: >
           further description of the award
       - name: left_reason
+      - name: left_reason_desc
```
pckhoi, line 306: What is the difference between `left_reason` and `left_reason_desc`? Would be nice if you leave a `description` for `left_reason_desc` at least.

ayyubibrahimi: Thanks.

ayyubibrahimi:
Ready for review.

pckhoi (approved):
LGTM

PR #319, opened 2022-06-19
Enhancement/post officer history reports

ayyubibrahimi:
@pckhoi The clean script for post officer history reports now reads from the output of the ner script. There is an issue with quality at the OCR stage. Nearly 50% of the rows are missing from the clean script when the input comes from the ner stage. Will continue to investigate.


pckhoi:
Thanks for letting me know. Feel free to add any parameter that you see fit to the ocr function.

ayyubibrahimi:
Closing this until the team decides on the future of our OCR workflow.

PR #311, opened 2022-06-05
Enhancement/ocr

ayyubibrahimi:
@pckhoi I've added an `ocr` and `ner` stage that is not integrated into deba, yet. Still working on the how. That said, before I went any further, I wanted to show you the following to confirm that integration is neccesary. I ask because I don't imagine having to ocr new documents more than twice a year, during which time ocr and ner may improve.

PR #320, opened 2022-06-21
Enhancement/pib_annual_reports

ayyubibrahimi:
@pckhoi Q for you. 

These PIB annual reports, from 2014-2019, contain a `tracking_id` which allows for them to be merged with our NOPD data. Because some `tracking_id`'s can be associated with 2 or more officers, the only way, I think, of matching them efficiently is to drop all of the `tracking_id`'s associated with 2 or more officers and subsequently match those `tracking_id`'s with our current NOPD data. 

Atm, our NOPD data is split into two scripts: `clean/cprr_new_orleans_pd_1931_2020.csv` (contains `tracking_id` and `officer_primary_key`) and `clean/pprr_new_orleans_ipm_iapro_1946_2018.csv` (contains `first_name`, `last_name` and `officer_primary_key`). 

What do you think is the best way to match?

### DIFF
--- a/dvc/data/raw/baker_pd.dvc
+++ b/dvc/data/raw/baker_pd.dvc
@@ -1,6 +1,6 @@
 wdir: ../../../data/raw
 outs:
-- md5: 50256a0059595867250fd90706d294c8.dir
-  size: 9842
+- md5: fd5f2ca26869fb62559bd6ccbb5d3a6b.dir
+  size: 7230
   nfiles: 3
   path: baker_pd
--- a/dvc/data/raw/post_officer_history.dvc
+++ b/dvc/data/raw/post_officer_history.dvc
@@ -1,0 +1,6 @@
+wdir: ../../../data/raw
+outs:
+- md5: b8bc7e6f1d623ce5f582a781dcabded2.dir
+  size: 256033
+  nfiles: 1
+  path: post_officer_history
--- a/fuse/new_orleans_pd.py
+++ b/fuse/new_orleans_pd.py
@@ -255,7 +255,7 @@ if __name__ == "__main__":
 
     complaints = fuse_cprr(cprr, actions, officer_number_dict)
     pib = fuse_pib(pib, officer_number_dict)
-    com = rearrange_allegation_columns(pd.concat([complaints, pib]))
+    com = rearrange_allegation_columns(pd.concat([complaints, pib]).drop_duplicates(subset=["allegation_uid"], keep="last"))
     personnel = fuse_personnel(
         pprr_ipm,
         lprr,
--- a/fuse/new_orleans_pd.py
+++ b/fuse/new_orleans_pd.py
@@ -32,7 +32,15 @@ def fuse_cprr(cprr, actions, officer_number_dict):
     )
     # cprr.loc[:, 'action'] = cprr.allegation_primary_key.map(
     #     lambda x: actions_dict.get(x, ''))
-    return rearrange_allegation_columns(cprr)
+    return cprr
+
+
+def fuse_pib(pib, officer_number_dict):
+    pib = float_to_int_str(cprr, ["officer_primary_key"])
+    pib.loc[:, "uid"] = cprr.officer_primary_key.map(
+        lambda x: officer_number_dict.get(x, "")
+    )
+    return pib
 
 
 def fuse_events(
@@ -242,9 +250,12 @@ if __name__ == "__main__":
     pprr_separations = pd.read_csv(
         deba.data("match/pprr_seps_new_orleans_pd_2018_2021.csv")
     )
+    pib = pd.read_csv(deba.data("match/cprr_new_orleans_pib_reports_2014_2019.csv"))
     brady = brady.loc[brady.agency == "New Orleans PD"]
 
     complaints = fuse_cprr(cprr, actions, officer_number_dict)
+    pib = fuse_pib(pib, officer_number_dict)
+    com = rearrange_allegation_columns(pd.concat([complaints, pib]))
     personnel = fuse_personnel(
         pprr_ipm,
         lprr,
@@ -274,7 +285,7 @@ if __name__ == "__main__":
     uof_df = rearrange_use_of_force(uof)
     brady_df = rearrange_brady_columns(brady)
     pclaims_df = rearrange_property_claims_columns(pd.concat([pclaims20, pclaims21]))
-    complaints.to_csv(deba.data("fuse/com_new_orleans_pd.csv"), index=False)
+    com.to_csv(deba.data("fuse/com_new_orleans_pd.csv"), index=False)
     personnel.to_csv(deba.data("fuse/per_new_orleans_pd.csv"), index=False)
     events_df.to_csv(deba.data("fuse/event_new_orleans_pd.csv"), index=False)
     lprr_df.to_csv(deba.data("fuse/app_new_orleans_csc.csv"), index=False)
--- a/fuse/new_orleans_pd.py
+++ b/fuse/new_orleans_pd.py
@@ -36,8 +36,8 @@ def fuse_cprr(cprr, actions, officer_number_dict):
 
 
 def fuse_pib(pib, officer_number_dict):
-    pib = float_to_int_str(cprr, ["officer_primary_key"])
-    pib.loc[:, "uid"] = cprr.officer_primary_key.map(
+    pib = float_to_int_str(pib, ["officer_primary_key"])
+    pib.loc[:, "uid"] = pib.officer_primary_key.map(
         lambda x: officer_number_dict.get(x, "")
     )
     return pib
--- a/fuse/new_orleans_pd.py
+++ b/fuse/new_orleans_pd.py
@@ -250,7 +250,7 @@ if __name__ == "__main__":
     pprr_separations = pd.read_csv(
         deba.data("match/pprr_seps_new_orleans_pd_2018_2021.csv")
     )
-    pib = pd.read_csv(deba.data("match/cprr_new_orleans_pib_reports_2014_2019.csv"))
+    pib = pd.read_csv(deba.data("match/cprr_new_orleans_pib_reports_2014_2020.csv"))
     brady = brady.loc[brady.agency == "New Orleans PD"]
 
     complaints = fuse_cprr(cprr, actions, officer_number_dict)
--- a/ner/nopd_pib_reports.py
+++ b/ner/nopd_pib_reports.py
@@ -16,46 +16,46 @@ def read_pdfs():
 def spacy_model(pdfs):
     ###  import labeled_data that has been exported from doccano
 
-    labeled_data = []
-    with open(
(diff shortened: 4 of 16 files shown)


ner/nopd_pib_reports.py
```diff
@@ -0,0 +1,85 @@
...
+    #         TRAINING_DATA, size=compounding(3.0, 1.5, 1.001)
+    #     ):
+    #         for text, annotations in batch:
+    #             doc = nlp.make_doc(text)
+    #             example = Example.from_dict(doc, annotations)
+    #             nlp.update([example], sgd=optimizer, losses=losses, drop=0.4)
+    #             print(losses)
+
+    # save model to disk:
+    # nlp.to_disk(deba.data("raw/new_orleans_pd/model/nopd_pib_reports.model"))
```
pckhoi, line 58: I see a lot of similarities in the commented out NER code compared to [ner/post_officer_history_reports.py](https://github.com/ppact/processing/blob/2cd172505cd308107d4dd8db4e83219d31f067cd/ner/post_officer_history_reports.py). Better extract them out to a library function.

And the training should be turned on/off via a make variable (e.g. `make TRAIN=y`) rather than commenting out code. What do you think?

pckhoi:
Just merge this with the CPRR using the `tracking_id`, it seems to be straightforward? Since there is probably no typo within the `tracking_id` (hopefully), you probably don't even need to use `datamatch`. But we should take care to include records from PIB that cannot be matched against CPRR, and records from CPRR that can't be matched against PPRR shouldn't have empty uid either.

It just seems very straightforward to me. Let me know if there is something I'm missing.

ner/nopd_pib_reports.py
ayyubibrahimi: Extracted the what was previously the `spacy_model` function into its sub-parts: `train_spacy_model` and `load_spacy_model`. Is commenting out the `train_spacy_model` function sufficient for this purpose? Please review.

ayyubibrahimi:
PIB records that can't be matched against CPRR should be dropped and records from CPRR that can't be matched against PPRR should be dropped. Is that not correct? There is no way to identify the officers if they do not have an officer_primary_key that maps onto the PPRR table. 

Yes, it is very straightforward at the moment. Sorry I was not very clear when I put forward the question yesterday. My concerns go beyond the moment, to the future when we decide to swap out these CPRR and PPRR tables that we received from IPM with CPRR and PPRR tables that we receive directly from the police agency. Not a concern for the moment. I'll resurface the thoughts at that point in time.

ayyubibrahimi:
No idea why the process-data check is failing. Am I missing something? In the last commit, I even changed the name, to act as a reset, but the process is still failing.

pckhoi:
> PIB records that can't be matched against CPRR should be dropped and records from CPRR that can't be matched against PPRR should be dropped. Is that not correct? There is no way to identify the officers if they do not have an officer_primary_key that maps onto the PPRR table.

Agreed

pckhoi:
Looks like the `data/meta` folder wasn't created. I think I need to handle the meta file creation with more sophistication.

pckhoi:
@ayyubibrahimi merging #321 (and making according changes) should fix the issue.

ayyubibrahimi:
@pckhoi There is a new error. Stumped by it.

pckhoi:
Looks like `pdf2image` isn't recorded in `requirements.txt`. My mistake. Can you append this line to `requirements.txt` and try again?
```
pdf2image==1.16.0
```

ayyubibrahimi:
I'm having additional issues in addition to the most recent process-data failure, namely my space has decreased from 20gb available to 0 while attempting to ocr, which means that I can no longer perform ocr on the pdfs. This has occured after I deleted nopd_pib_reports from the `ocr_cache` in the interest of experimenting with ocr parameters. So far I've had no luck in my search for solutions.

pckhoi:
The only thing that I'm aware of being generated is the OCR text dumped as JSON files in the `ocr_cache` folder. Can you check the size of that folder? Beside that `pytesseract` could generate some artifacts during OCR.

ayyubibrahimi:
Cache folder looks good. I'll look more into pytesseract. Thanks!

ayyubibrahimi:
@pckhoi I believe the issue is that `pytesseract` is reading the pdf into memory at once. The size of each pdf is roughly 669kb. I receive the message below when I attempt to OCR, and my space goes from x gb to 0 bytes. 

I was able to recover roughly 4gb of memory when I uninstalled and re-installed `pytesseract` but my space disappeared again when I re-attempted to OCR the pib pdfs. 

Addtionally, I am able to successfully OCR the post history pdfs from PR #311, which are between 70-250kb in size.

If this sounds correct to you, do you advise that I look into re-writing the ocr function, or should I split the pdfs into multiple pdfs to decrease the size? 

```
OCR pdf:   0%|          | 0/6 [00:14<?, ?it/s]
    Traceback (most recent call last):
      File "/mnt/c/Users/ayyubi/lleads/processing/ocr/nopd_pib_reports.py", line 21, in <module>
        df = process_all_pdfs()
      File "/mnt/c/Users/ayyubi/lleads/processing/ocr/nopd_pib_reports.py", line 14, in process_all_pdfs
        pd.read_csv(deba.data("meta/nopd_pib_report_files_2014_2019.csv"))
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/pandas/core/generic.py", line 5406, in pipe
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/pandas/core/common.py", line 440, in pipe
      File "/mnt/c/Users/ayyubi/lleads/processing/lib/ocr.py", line 75, in process_pdf
        txt = pytesseract.image_to_string(image)
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/pytesseract/pytesseract.py", line 416, in image_to_string
        return {
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/pytesseract/pytesseract.py", line 419, in <lambda>
        Output.STRING: lambda: run_and_get_output(*args),
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/pytesseract/pytesseract.py", line 275, in run_and_get_output
        with save(image) as (temp_name, input_filename):
      File "/usr/lib/python3.9/contextlib.py", line 117, in __enter__
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/pytesseract/pytesseract.py", line 198, in save
        image.save(input_file_name, format=image.format)
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/PIL/Image.py", line 2300, in save
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/PIL/PpmImagePlugin.py", line 182, in _save
      File "/home/ayyubibrahimi/.local/lib/python3.9/site-packages/PIL/ImageFile.py", line 524, in _save
    OSError: [Errno 5] Input/output error
```

And, finally, re the process-data failure, am I correct in that I also need to add `poppler` to `requirements.txt`

pckhoi:
Poppler and Tesseract are C++ programs. I added a step to install them for Ubuntu.

Reading the `process_pdf` function again, it seems to output images to a temporary folder. Sorry for missing that. I have pushed a new version with the argument `output_images_to_tempdir` that when set to False will keep all images in memory. This could potentially get the process OOM-killed instead. You can give it a try and let me know how it goes:

```python
.pipe(process_pdf, dir_name, output_images_to_tempdir=False)
```

pckhoi:
Looks like OCR and NER steps completed successfully.

ayyubibrahimi:
Sounds good. I will re-try and let you know how it goes. Thanks!

pckhoi:
No worries. I would suggest utilizing LLEAD's Kubernetes cluster for OCR but I think getting more disk space on your end would be a more permanent solution.

ayyubibrahimi:
@pckhoi I'm no longer able to push to the Google `excellent-zoo-300106` bucket because I no longer have access, but it also seems to no longer exists. I just forwarded you an email that I missed back in May. Please advise.

pckhoi:
`excellent-zoo-300106` is the Google Cloud project id. It certainly still exists. Please login again with this:
```
gcloud auth login
dvc remote modify --local gcs credentialpath ~/.config/gcloud/legacy_credentials/<your email>/adc.json
```

ayyubibrahimi:
Sorry. Stupid mistake. I was looking into the wrong account. Thanks. This is ready for review the whenever you have the time.

dvc/data/raw/post_officer_history.dvc
```diff
@@ -0,0 +1,6 @@
+wdir: ../../../data/raw
+outs:
+- md5: b8bc7e6f1d623ce5f582a781dcabded2.dir
+  size: 256033
+  nfiles: 1
+  path: post_officer_history
```
pckhoi, line 6: This file seems to belong to another PR. Any idea why it made it into this branch?

dvc/data/raw/baker_pd.dvc
```diff
@@ -1,6 +1,6 @@
 wdir: ../../../data/raw
 outs:
-- md5: 50256a0059595867250fd90706d294c8.dir
-  size: 9842
+- md5: fd5f2ca26869fb62559bd6ccbb5d3a6b.dir
+  size: 7230
   nfiles: 3
   path: baker_pd
```
pckhoi, line 6: Same thing here. We should take care to not commit irrelevant changes.

fuse/new_orleans_pd.py
```diff
@@ -32,7 +32,15 @@ def fuse_cprr(cprr, actions, officer_number_dict):
...
     #     lambda x: actions_dict.get(x, ''))
-    return rearrange_allegation_columns(cprr)
+    return cprr
+
+
+def fuse_pib(pib, officer_number_dict):
+    pib = float_to_int_str(cprr, ["officer_primary_key"])
+    pib.loc[:, "uid"] = cprr.officer_primary_key.map(
+        lambda x: officer_number_dict.get(x, "")
+    )
```
pckhoi, line 42: `cprr` needs to be passed in explicitly as a function argument.

pckhoi:
How come there are no change in the `allegation` branch? I see that the `fuse/new_orleans_pd.py` script append PIB data to `data/fuse/com_new_orleans_pd.csv`. So there should be some changes. Can you investigate @ayyubibrahimi ?

fuse/new_orleans_pd.py
ayyubibrahimi: Thanks

dvc/data/raw/post_officer_history.dvc
ayyubibrahimi: I believe that this is from my experimenting with OCR when it was failing to work. Sorry for committing.

dvc/data/raw/baker_pd.dvc
ayyubibrahimi: I could not say. I'll be more careful to not commit irrelevant changes. Thanks.

ayyubibrahimi:
@pckhoi Figured it out. Ready for review.

pckhoi (approved):
LGTM

PR #321, opened 2022-06-23
Upgrade deba to 0.1.6

pckhoi:
Deba version 0.1.6 introduces `references` pattern which are like `prerequisites` except they are assumed to be version controlled by Git and not contained in `dataDir`.


ayyubibrahimi (approved):
Awesome.

PR #326, opened 2022-08-24
bugs/post_factory_reset

ayyubibrahimi:
- I removed the `lprr_baton_rouge_fpcsb_1992_2012.csv` entirely from the pipeline after further review. The original pdf itself is riddled with inconsistencies that are nearly impossible to decipher, which is why the clean script that you originally produced took so much work. Even with that work, it's not certain that the data is accurate only because the data in the pdf itself doesn't make much sense. My two potential solutions are: 1) I'm going to attempt to re-request that document from the agency itself; 2) see if the output is improved after we introduce the new OCR pipeline. Thoughts?


ayyubibrahimi:
Re the process-data fail. Looks like it's not accepting my 4.3 version of `make`? 

Output of when I run `make -version` is below:

```
ayyubi@DESKTOP-BLLRMEO:/mnt/c/Users/PC/Desktop/lleader/processing$ make -version
GNU Make 4.3
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

pckhoi:
Your local version of Make has nothing to do with the Make installation in Github runner. Looks like the mirror link that we relied on is dead.

pckhoi:
@ayyubibrahimi this PR should fix all Github runner issues: #327

pckhoi:
Regarding the lprr script, yeah, I don't like black magic anywhere in the code. I guess any of those 2 solutions would be better than keeping that script as is.

ayyubibrahimi:
Cool. This is ready for review at your convenience.

pckhoi (approved):
Lgtm

PR #327, opened 2022-08-25
Using custom Github runner image

pckhoi:
- Install Python, Make, and dependencies in Docker image.
- Reduce job run time.
- **IMPORTANT**: After this change, we need to manually run `gh_k8s_runner/install.sh <gcloud_project_id>` whenever we update `requirements.txt` so that the Github runner image is also updated.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*
+
+!/requirements.txt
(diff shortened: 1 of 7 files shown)


ayyubibrahimi (approved):
Awesome.

PR #329, opened 2022-08-29
Enhancement/nopd_appeal_transcripts

ayyubibrahimi:


ayyubibrahimi:
@pckhoi - I've added the meta data produced in the `meta` stage to the `datavalid.yml` file for the `appeal` schema. 

With the intention that this would allow for the dev team to append documents to LLEAD when a corresponding document exist. This is possible I think, for example, with `fuse/app_new_orleans_pd.csv`.

Does that make sense to you?

pckhoi:
Are you saying that the dev team should be allowed to add documents into LLEAD WITHOUT going through this pipeline? If that's what you mean then my intuition says no. And where is the `fuse/app_new_orleans_pd.csv` file? Doesn't seem to exist in this PR.

datavalid.yml
```diff
@@ -362,6 +362,14 @@ schemas:
...
       - name: agency
       - name: motions
+      - name: md5
+      - name: filepath
+      - name: filesha1
+      - name: fileid
+      - name: filetype
+      - name: fn
+      - name: file_category
+      - name: text
```
pckhoi, line 372: My main question would be whether we are dealing with 1 file per row or not. Doing this means that you can no longer mix and match data from different files to produce a single output anymore.

pckhoi: And I still don't understand how this helps adding document. Could you explain more?

ayyubibrahimi:
Re: "Are you saying that the dev team should be allowed to add documents into LLEAD WITHOUT going through this pipeline? If that's what you mean then my intuition says no. And where is the fuse/app_new_orleans_pd.csv file? Doesn't seem to exist in this PR."

Apologies. I meant `fuse/app_new_orleans_csc.csv`

Not at all. I meant the opposite. I'm attempting to create a way for the dev team to add documents to LLEAD through this pipeline. At the moment, I'm not sure that there is anyway for them to do this after documents, such as these appeal hearing transcripts, have been processed through the meta, `ocr`, `ner`, `clean`, `match`, and `fuse` stages.

datavalid.yml
ayyubibrahimi: It was my understanding that these metadata would allow for the dev team to add docs to LLEAD. 

It does look as if we are no longer dealing with 1 file per row. Please advise.

pckhoi: These metadata are only there to assists automatic tools and libraries, mainly for OCR and features extraction, not for human nor downstream jobs to consume. Please follow the KISS principle. If it's not immediately necessary, don't put it in.

ayyubibrahimi: Got it. I will make these changes. 

What do you advise is the best way for me on the processing side to assist the dev team with adding these transcripts to LLEAD?

pckhoi: Please clarify for me, in what ways do you think adding transcripts to LLEAD is hard right now?

Anyone who want to inherit and extend this pipeline is expected to run the pipeline, inspect the output, and read the code and/or documentation to understand what is going on. In that light, the best way for you to help is to document the process, making clear and clarify what is not easy to understand. Also simplifying the process more and more, not complicating it more. Simplifying has also been the main idea behind my various libraries.

Also we don't want to over-simplify the process. We shouldn't try to sell a difficult process as easy and trivial. It's a hard balance to achieve.

ayyubibrahimi: My initial understanding of how the dev team added docs to LLEAD was based on Tarak's documents [repo](https://hub.wrgl.co/@ipno/r/data/refs/heads/documents) on WrglHub. I took inspiration from the many metadata fields.   

I believe that I misunderstood the dev team's processing for adding docs. In the latest commit, I have removed all metadata from the tables.

pckhoi: I see where you're coming from. To be honest, I don't know what is the full purpose of the meta fields either. Maybe it is just good practice to produce metadata for pdfs that will be OCRed. They are certainly useful for automated tools such as the (unfinished) Spot Check vscode extension. But Tarak probably knows more than me.

And I might be wrong too. I'm simply coming from a purely utilitarian perspective. If something is not immediately useful, discard.

ayyubibrahimi: Makes sense. 

With the metadata fields discarded in the latest commit, this PR is now ready for review.

pckhoi:
This branch seems to run fine at least. However there are too many Git conflicts. If you are merely switching between branches that you yourself made then there should be no conflict whatsoever. Forget about Kubernetes or Github action, you cannot be entrusted with handling infrastructure as long as mistakes like this can still happen. Please focus on acquiring solid Git skill next.

ayyubibrahimi:
Got it. I will focus switch my focus to Git. That said, these conflicting files seem to be consistent with what PR's have been merged since this branch passed its checks 11 day ago.

pckhoi:
I see. Conflicts could be present simply because a branch is just old. Apologies.

PR #415, opened 2022-11-11
Enhancement/terrebonne uof

ayyubibrahimi:


pckhoi:
Looks like the process was OOM-killed. I guess there was a lot of data? I would love to fix this permanently but it's not easy to control how much memory badger consume. Just restart the processing until it succeed. Badger should have saved some data along the way so it's not likely to run out of memory again.

ayyubibrahimi:
Got it. Thanks!

PR #480, opened 2023-03-14
bug/updated apply transaction and make

ayyubibrahimi:


.github/workflows/apply_transaction.yml
rajivsinclair, line 1: @ayyubibrahimi the schema validator was added by Bao but then your most recent commits have removed it. Was this intentional or just an out-of-sync commit?

rajivsinclair: (originally added [here](https://github.com/ipno-llead/processing/commit/2363d4275ecd3054419ee9d180c8ff9b1128a03f))

ayyubibrahimi: Misunderstanding on my part. I've just opened a PR to replace it.

PR #506, opened 2023-04-06
Copied the work from the fork repo to the original repo in order to u…

hungEA:
**Title:** [248 - Develop data validator](https://www.pivotaltracker.com/story/show/184613248)

**Description:**

The data-validator program will be able to run validation check for the required csv files that would be push to WRGL before being imported to BE database. The process will be applied to the following data

- [x] agency_reference_list
- [x] personnel
- [x] allegation
- [x] appeals
- [x] use_of_force
- [x] citizens
- [x] event
- [x] person
- [x] documents

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 numpy==1.23.4
 pandas==1.4.2
-python-Levenshtein==0.12.0
+levenshtein==0.20.9
 Unidecode==1.1.2
 datamatch==0.1.15
 datavalid==0.3.6
@@ -22,3 +22,5 @@ pypdfium2==2.11.0
 pathspec==0.9.0
 dropbox==11.34.0
 vscode-spot-check==0.0.1
+psycopg2-binary==2.9.2
+slack_sdk==3.19.5
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 numpy==1.23.4
 pandas==1.4.2
-levenshtein==0.20.9
+python-Levenshtein==0.12.0
 Unidecode==1.1.2
 datamatch==0.1.15
 datavalid==0.3.6
--- a/.github/workflows/process_data.yml
+++ b/.github/workflows/process_data.yml
@@ -6,6 +6,25 @@ on:
 jobs:
   process-data:
     runs-on: cdi-runner
+    # Service containers to run with `container-job`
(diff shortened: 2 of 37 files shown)


baoea:
Hi @ayyubibrahimi. Could you explain us why this process fails? We need to pass this step to test our validator. Thanks!

ayyubibrahimi:
I'll review.

ayyubibrahimi:
I will re-write this function.

ayyubibrahimi:
These bugs are weird. I can't reproduce them on my local when I run `make`, `make` ran fine with the process-data step for my last PR (the PR that you just merged into this branch), and no arguments are being passed to the similarity function. Can you please advise @pckhoi.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 numpy==1.23.4
 pandas==1.4.2
-python-Levenshtein==0.12.0
+levenshtein==0.20.9
```
pckhoi, line 3: @ayyubibrahimi this is the culprit. `datamatch` is dependent on python-Levenshtein v0.12.0. I wonder why this dependency is updated in the first place.

hungEA: If we don't update that dependency, we can't build our docker image. Have you recently tried to rebuild the docker image with `python-Levenshtein==0.12.0`?

pckhoi: Actually they seem to be 2 separate packages, so they can probably coexist. Can you try putting back in the python-levenshtein package?

pckhoi: If that doesn't work I'll update datamatch. It should work because both packages seem to be just different versions.

ayyubibrahimi: Thanks Khoi. I'm not near my laptop atm but I'll update it ASAP.

hungEA: I reverted it back to the original version. I also had to clean up the cache in the docker builder. It's successfully built the docker image.

hungEA:
@ayyubibrahimi @pckhoi our data-validator processes have successfully passed. However, we are not sure why the `commit changes in transaction` step failed as we did not touch this section at all. Can you have a look please?

ayyubibrahimi:
Awesome. From what I understand this is a issue with the server connection. Please re-run the failed jobs and it should eventually process successfully.

PR #513, opened 2023-04-28
Enhancement/data validator update

hungEA:
The following changes to `data-validator` are applied:

- [x] Add new attribute coaccusal to allegation record
- [x] Add validation check for brady data
- [x] Build relationship between event and brady

### DIFF
--- a/fuse_agency/all_agency.py
+++ b/fuse_agency/all_agency.py
@@ -374,6 +374,7 @@ if __name__ == "__main__":
     per_df = fuse_personnel()
     ensure_uid_unique(per_df, "uid")
     event_df = fuse_event()
+    ensure_uid_unique(event_df, "event_uid")
     allegation_df = fuse_allegation()
     ensure_uid_unique(allegation_df, "allegation_uid")
     uof_df = fuse_use_of_force()
--- a/fuse_agency/all_agency.py
+++ b/fuse_agency/all_agency.py
@@ -374,7 +374,6 @@ if __name__ == "__main__":
     per_df = fuse_personnel()
     ensure_uid_unique(per_df, "uid")
     event_df = fuse_event()
-    ensure_uid_unique(event_df, "event_uid")
     allegation_df = fuse_allegation()
     ensure_uid_unique(allegation_df, "allegation_uid")
     uof_df = fuse_use_of_force()
--- a/data-validator/allegation_importer.py
+++ b/data-validator/allegation_importer.py
@@ -122,7 +122,7 @@ def run(db_con, allegation_df, allegation_cols):
             FROM stdin WITH CSV HEADER
             DELIMITER as ','
         """,
-        file=open('complaints.csv','r'),
+        file=open('complaints.csv', 'r'),
     )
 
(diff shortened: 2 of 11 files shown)


hungEA:
@ayyubibrahimi Could you please check the process of generating event data? There seems to be a unique constraint for event_uid.

ayyubibrahimi:
@baoea there is a unique constraint built into the `Builder` class for the events module lib/events.py line 419, but I don't believe that you are accessing this module. 

This process-data error is strange. In this processing repo, an error raised when there are duplicate `event_uid` at the fuse_agency stage, which prevents the processing repo from completing successfully. Additionally, I dropped any duplicate `event_uid` (just in case) before the final `event` table is output, but this process-data failure persists. Can you say more about which iteration of the `event` table is being fetched prior to the error?

hungEA:
@ayyubibrahimi  We're certainly not using your lib/events.py. The one that raised the error was in our validator `https://github.com/ipno-llead/processing/blob/enhancement/data-validator-update/data-validator/event_importer.py`. The issue here is that our validator did not add extra data to your processed event.csv, but when we imported the `event.csv` in the `fuse` folder to a temporary database, it threw error about unique constraint violation.

```
File "/runner/_work/processing/processing/data-validator/data_validator.py", line 86, in run_validator
    module.run(conn, df, be_cols)
  File "/runner/_work/processing/processing/data-validator/event_importer.py", line 167, in run
    cursor.copy_expert(
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "officers_event_event_uid_acf5b8ca_uniq"
DETAIL:  Key (event_uid)=(76a723fd1140db6abd7f0db0f53d43f2) already exists.
CONTEXT:  COPY officers_event, line 4617
```

So please double-check the output of the data in the fuse/event.csv for this `event_uid=76a723fd1140db6abd7f0db0f53d43f2` as we have no way to check it before it's push to WRGL.

ayyubibrahimi:
Hi @hungEA, both on my local and on wrgl there is only one entry for the `event_uid=76a723fd1140db6abd7f0db0f53d43f2`. I updated the fuse stage [here](https://github.com/ipno-llead/processing/pull/513/commits/7e8fd5412468a140ac1a4141a1334631f3a93bb6) so that duplicate `event_uid` are dropped before the `fuse/event.csv` table is generated, but as we see the error is still thrown which is why I asked about where this table was being fetched from, because the current code does not allow for duplicates.

fuse_agency/all_agency.py
```diff
@@ -374,7 +374,6 @@ def fuse_agency_lists():
     per_df = fuse_personnel()
     ensure_uid_unique(per_df, "uid")
     event_df = fuse_event()
-    ensure_uid_unique(event_df, "event_uid")
```
hungEA, line 377: @ayyubibrahimi I wonder why you remove this call?

hungEA:
@ayyubibrahimi We have fixed our `brady_uid` constraint in the BE schema. Please have a look at the error of duplicated `brady_uid` in `brady` table.

PR #566, opened 2024-04-16
[c] Clean up WRGL

ea-open-source:
## Description
This PR provides the clean up WRGL in the Github workflows.

Since we are not so familiar with the internal logic of the processing workflow, we really need your thorough review of this PR @ayyubibrahimi


.github/workflows/process_data.yml
```diff
@@ -41,12 +41,8 @@ jobs:
           dvc pull -v -f
       - name: wrgl pull
         run: |
-          wrgl config set --global user.email "gh-action@publicdata.works"
-          wrgl config set --global user.name "Github Action"
           mkdir /runner/_work/data/.wrgl || true
           cp .wrgl/config.yaml /runner/_work/data/.wrgl/
```
ea-open-source, line 45: We see that the `WRGL_FLAGS` is being used on [line 54](https://github.com/ipno-llead/processing/pull/566/files#diff-8435a45930c33e0ed8f774d398d31b8daabe3133d453ec7a4bb6ea50f8f4f8f9L54), so we are not sure that these lines can be removed.

PR #575, opened 2024-10-25
store CSV files for each agency and write their URLs to the db

rajivsinclair:
This commit introduces two key workflow changes to handle agency-specific CSV files:

1. Enhanced upload_csv_to_google_cloud.yml:
   - Added support for uploading agency-specific CSVs from fuse_agency directory
   - Generates a JSON file (csv_urls.json) mapping filenames to their public URLs
   - Sets appropriate public access permissions for uploaded files

2. Added new workflow update_wrgl_urls.yml:
   - Automatically triggers after successful CSV upload
   - Securely connects to Cloud SQL using Cloud SQL Proxy
   - Downloads the generated csv_urls.json file
   - Updates departments_wrglfile table in PostgreSQL database
   - Maps CSV filenames to agency slugs (e.g., per_new_orleans_pd.csv → new-orleans-pd)
   - Updates both url and download_url fields for each agency

Required Secrets:
- GCP_SA_KEY: Google Cloud service account credentials
- GCP_PROJECT_ID: Google Cloud project identifier
- CLOUD_SQL_DATABASE: Cloud SQL instance name
- POSTGRES_DB, POSTGRES_USER, POSTGRES_PASSWORD: Database credentials

This change ensures that agency-specific CSV files are automatically uploaded to Google Cloud Storage and their URLs are properly updated in the [LLEAD Django](https://github.com/publicdataworks/llead-backend) application's database, maintaining synchronization between data processing and the frontend application.

### DIFF
--- a/.github/workflows/update_wrgl_urls.yml
+++ b/.github/workflows/update_wrgl_urls.yml
@@ -27,19 +27,24 @@ jobs:
 
       - name: Download and verify csv_urls.json
         run: |
+          # Get the folder name using the same pattern as the upload workflow
           FOLDER_NAME="data-${{ github.event.workflow_run.id }}-${{ github.event.workflow_run.run_attempt }}"
+          
           echo "Checking for csv_urls.json in bucket..."
           if ! gsutil -q stat "gs://${{ vars.DATA_BUCKET }}/$FOLDER_NAME/csv_urls.json"; then
             echo "Error: csv_urls.json not found in cloud storage"
             exit 1
           fi
+          
           echo "Downloading csv_urls.json..."
           gsutil cp "gs://${{ vars.DATA_BUCKET }}/$FOLDER_NAME/csv_urls.json" ./csv_urls.json
+          
           echo "Verifying JSON format..."
           if ! jq empty csv_urls.json; then
             echo "Error: Invalid JSON format in csv_urls.json"
             exit 1
           fi
+          
           echo "Verifying JSON content..."
           URL_COUNT=$(jq 'length' csv_urls.json)
           echo "Found $URL_COUNT URL entries in csv_urls.json"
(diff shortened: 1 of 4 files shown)


rajivsinclair:
@ayyubibrahimi how does this look to you?

.github/workflows/upload_csv_to_google_cloud.yml
```diff
@@ -101,6 +101,37 @@ jobs:
         destination: '${{ vars.DATA_BUCKET }}/${{ env.FOLDER_NAME }}'
         process_gcloudignore: false
 
+    # Upload agency-specific input CSVs and save URLs
+    - id: 'upload-agency-csvs'
+      uses: 'google-github-actions/upload-cloud-storage@v2'
+      with:
+        path: '${{ vars.DATA_DIR }}/fuse_agency'
```
rajivsinclair, line 108: @ayyubibrahimi do you know how we should correctly reference the location of the ephemeral CSV files that were generated in the previous steps of the GitHub Actions?

.github/workflows/upload_csv_to_google_cloud.yml
```diff
@@ -101,6 +101,75 @@ jobs:
...
+        echo "$UPLOADED_FILES"
+        
+        # Generate URLs for verified files
+        echo "Generating public URLs..."
+        echo "$UPLOADED_FILES" | while read -r file; do
+          if [ ! -z "$file" ]; then
+            filename=$(basename "$file")
+            filetype=${filename%%_*}  # Extract type (per, com, uof, etc.)
+            echo "Processing $filename (type: $filetype)..."
+            echo "  \"${filename}\": \"https://storage.googleapis.com/${file#gs://}\"," >> csv_urls.json
```
tarakc02, line 149: What does the `#gs:` do here? pretty nitpicky, just making sure i follow

.github/workflows/update_wrgl_urls.yml
```diff
@@ -0,0 +1,172 @@
...
+      - name: Update Database
+        env:
+          POSTGRES_DB: ${{ secrets.POSTGRES_DB }}
+          POSTGRES_USER: ${{ secrets.POSTGRES_USER }}
+          POSTGRES_PASSWORD: ${{ secrets.POSTGRES_PASSWORD }}
+        run: |
+          sudo apt-get update && sudo apt-get install -y python3-pip
+          pip install psycopg2-binary
+          cat << 'EOF' > update_wrgl.py
+          import psycopg2
```
tarakc02, line 60: probably a really obvious answer that i'm missing, but why do we construct this python script on the fly this way rather than just having it as part of the regular code repo? so that this step would still do the package installs, then just call `python3 path/to/update_wrgl.py`?

.github/workflows/update_wrgl_urls.yml
```diff
@@ -0,0 +1,172 @@
...
+              if match:
+                  return match.group(1)
+              return None
+
+          def update_wrgl_urls():
+              success_count = 0
+              error_count = 0
+              skipped_count = 0
+              print("Starting database update process...")
+              try:
```
tarakc02, line 82: i struggled a bit following the triply-nested try/except blocks. at least the first one seems like it could be pulled out -- if we can't establish a db connection then nothing else is going to happen anyways.

.github/workflows/update_wrgl_urls.yml
```diff
@@ -0,0 +1,172 @@
...
+                          print("\nFiles processed by type:")
+                          for file_type, count in type_stats.items():
+                              print(f"- {file_type}: {count} files")
+                          if error_count > 0:
+                              raise Exception(f"Encountered {error_count} errors during update")
+                  except json.JSONDecodeError as e:
+                      raise Exception(f"Failed to parse csv_urls.json: {str(e)}")
+              except Exception as e:
+                  print(f"ERROR: {str(e)}")
+                  raise
```
tarakc02, line 163: not raising anything -- maybe `raise e`?

tarakc02 (approved):
left some questions/notes in the review comments, thanks!

.github/workflows/update_wrgl_urls.yml
rajivsinclair: thanks @tarakc02 that's a good point. this happened from an instinct to minimize the number of new files. you're absolutely right that this script should just be in the repo and then we can simply reference it from the yaml. here's the update: https://github.com/ipno-llead/processing/pull/575/commits/db3f8ea9c3b8b5d1cce63603d559422370fec9ec

.github/workflows/update_wrgl_urls.yml
rajivsinclair: thanks https://github.com/ipno-llead/processing/pull/575/commits/6dfd15edba1833c8e703790323d3fb3219a332dc

.github/workflows/update_wrgl_urls.yml
tarakc02: (by "pulled out") i just mean into its own block

.github/workflows/upload_csv_to_google_cloud.yml
rajivsinclair: strips the gs:// prefix from the file variable – this is to convert the Google Cloud Storage URI into a standard HTTPS URL by removing the gs:// prefix and appending the remaining path to https://storage.googleapis.com/.

.github/workflows/update_wrgl_urls.yml
rajivsinclair: yea, you're right. cleaned it up a bit here: https://github.com/ipno-llead/processing/pull/575/commits/e88fff368798bf154c0cf987ca9579d25bc69561

.github/workflows/upload_csv_to_google_cloud.yml
ayyubibrahimi: I believe that this right @rajivsinclair

tarakc02 (approved):

PR #604, opened 2025-05-01
Processing (clean/match/fuse_agency) scripts for natchitoches so 

esmel200:
Adding cprr 2022-2025 processing scripts, minor revisions to cprr 2018-2021 cleaning function for Natchitoches SO.

### DIFF
--- a/clean/natchitoches_so_cprr.py
+++ b/clean/natchitoches_so_cprr.py
@@ -89,44 +89,43 @@ def clean():
     )
     return df
 
-RANK_MAP = {
-    "dy": "deputy",
-    "det": "detective",
-    "lt": "lieutenant",
-    "sgt": "sergeant",
-    "cpt": "captain",
-    "mjr": "major"
-}
 
 def split_multiple_officers(df):
+    RANK_MAP = {
+        "dy": "deputy",
+        "det": "detective",
+        "lt": "lieutenant",
+        "sgt": "sergeant",
+        "cpt": "captain",
+        "mjr": "major"
+    }
+
     rows = []
 
     for _, row in df.iterrows():
         deputy_str = row["deputy_s_name"].strip().lower()
 
-        # Handle organizational/collective entries like "NPSO Task Force"
-        if deputy_str.startswith("npso"):
-            new_row = row.copy()
-            new_row["rank_desc"] = deputy_str  # Full label like "npso task force"
-            new_row["first_name"] = "none"
-            new_row["middle_name"] = "none"
-            new_row["last_name"] = "none"
-            rows.append(new_row)
+        if "task force" in deputy_str or "npso general" in deputy_str:
             continue
 
-        # Normal individual officers with rank + name
         officers = re.findall(r"(\w+)\.?\s+([\w'-]+)", deputy_str)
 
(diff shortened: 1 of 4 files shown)


clean/natchitoches_so_cprr.py
```diff
@@ -1,94 +1,171 @@
...
+}
+
+def split_multiple_officers(df):
+    rows = []
+
+    for _, row in df.iterrows():
+        deputy_str = row["deputy_s_name"].strip().lower()
+
+        # Handle organizational/collective entries like "NPSO Task Force"
+        if deputy_str.startswith("npso"):
```
ayyubibrahimi, line 108: We should drop rows where there is no valid name

clean/natchitoches_so_cprr.py
```diff
@@ -1,94 +1,171 @@
...
+        pd.read_csv(deba.data("raw/natchitoches_so/cprr_natchitoches_so_2022_2025.csv"))
+        .pipe(clean_column_names)
+        .rename(columns={"investigation_notes": "investigation_desc"})
+        .pipe(strip_leading_commas)
+        .pipe(format_dates)
+        .pipe(clean_allegation)
+        .pipe(clean_disposition)
+        .pipe(clean_action)
+        .pipe(split_multiple_officers)
+        #.pipe(clean_split_names)
```
ayyubibrahimi, line 155: We should remove any lines of code that are commented out

clean/natchitoches_so_cprr.py
```diff
@@ -1,94 +1,171 @@
...
+        .pipe(clean_split_names)
+        .pipe(clean_dates, ["receive_date", "disposition_date"])
+        .pipe(standardize_desc_cols, ["allegation", "investigation_desc"])
+        .pipe(set_values, {"agency": "natchitoches-so"})
+        .pipe(gen_uid, ["first_name", "last_name", "agency"])
+        .pipe(gen_uid, ["uid", "receive_year", "allegation", "disposition", "action"], "allegation_uid")
+    )
+    return df
+
+RANK_MAP = {
```
ayyubibrahimi, line 92: Can you define the rank_map within its relevant function

clean/natchitoches_so_cprr.py
esmel200: Thank you! I meant to ask you about the complaints w no name. Will put in these changes now and check for more.

clean/natchitoches_so_cprr.py
```diff
@@ -1,94 +1,171 @@
...
+            new_row["first_name"] = "none"
+            new_row["middle_name"] = "none"
+            new_row["last_name"] = "none"
+            rows.append(new_row)
+            continue
+
+        # Normal individual officers with rank + name
+        officers = re.findall(r"(\w+)\.?\s+([\w'-]+)", deputy_str)
+
+        for rank_abbr, last_name in officers:
```
ayyubibrahimi, line 120: Why are you setting first_name to none?

esmel200: There were no names for two complaints. The "officer name" in the raw data was NPSO task force and NPSO (general). Those rows should be dropped instead, correct?

PR #607, opened 2025-05-06
updated nopd clean/match/fuse scripts to include nopd cprr 2024 by co…

esmel200:
…ncatenating clean cprr 2024 data w previous concatenated clean cprr csv w 2021-23 and 2025 nopd cprr data


clean/new_orleans_pd_cprr.py
```diff
@@ -332,6 +332,42 @@ def clean25():
...
+        .pipe(clean_column_names)
+        .rename(
+            columns={
+                "fdi": "tracking_id_og",
+                "incident_type": "complainant_type",
+                "accused_employee_id": "employee_id",
+                "relevant_policy": "allegation_desc"
+            }
+        )
+        .pipe(clean_tracking_id_23)
```
ayyubibrahimi, line 350: Did you check all of the values in these rows manually in the terminal to make sure that the functions for `23` work for `24`?

clean/new_orleans_pd_cprr.py
```diff
@@ -370,5 +406,6 @@ def clean23():
     dfa = clean()
     dfb = clean23()
     dfc = clean25()
-    df = pd.concat([dfa, dfb, dfc])
-    df.to_csv(deba.data("clean/cprr_new_orleans_pd_2021_2023.csv"), index=False)
\ No newline at end of file
+    dfd = clean24()
+    df = pd.concat([dfa, dfb, dfc,dfd])
+    df.to_csv(deba.data("clean/cprr_new_orleans_pd_2021_2025.csv"), index=False)
```
ayyubibrahimi, line 411: The years look wrong. Think this should be something like 2005_2025

clean/new_orleans_pd_cprr.py
esmel200: I just did that and checked unique values for any cols that I used previous functions. It all looks clean but do you want me to try to write a function to double check? Fixed dates so will push again.

PR #609, opened 2025-05-06
iss608 adding UID unittest for common last names match script function

esmel200:
UID unittest for common last names match script function: checks that UID does not change for common last names in post data when matching it to csvs with only last_name data.

### DIFF
--- a/lib/uid_test.py
+++ b/lib/uid_test.py
@@ -31,3 +31,69 @@ class UIDTestCase(unittest.TestCase):
         ensure_uid_unique(dfa, ["a"])
         with self.assertRaisesRegex(Exception, r"rows are not unique:"):
             ensure_uid_unique(dfb, ["a"])
+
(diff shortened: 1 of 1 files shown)


lib/uid_test.py
```diff
@@ -31,3 +31,69 @@ def test_ensure_uid_unique(self):
         ensure_uid_unique(dfa, ["a"])
         with self.assertRaisesRegex(Exception, r"rows are not unique:"):
             ensure_uid_unique(dfb, ["a"])
+
+class TestCprrPostUidMatching(unittest.TestCase):
+    def test_common_last_name_uid_unchanged(self):
+        natchitoches_df = pd.DataFrame({
```
ayyubibrahimi, line 37: Can you extend this to include examples where we have first names as well for some people? Where the goal is to ensure that we skip entity resolution for persons that A) have no first name and B) a common last name

PR #612, opened 2025-05-21
El covington pd pprr 2002 2013

esmel200:
covington pprr cleaning scripts. I did not extract events in fuse_agency because the data only has data_production_year. Lmk your thoughts on how to deal w this issue. Ended up leaving extract events out of fuse_agency because baton rouge pd pprr21 had similar dataset, and events were not extracted.


ayyubibrahimi:
We shouldn't add anything that doesn't have an event (hire date, separation date, etc)

